PR #2, opened 2016-09-27
Canavandl/fix bokeh element ids

canavandl:
fix ids

### DIFF
--- a/src/widget.ts
+++ b/src/widget.ts
@@ -18,19 +18,22 @@ class DistributedUIElement extends Widget {
   constructor(script: any) {
     super();
 
+    // store bokeh model id as private attr for access in onResize eventing
+    this._bokeh_id = script["data-bokeh-model-id"]
+
     let tag = document.createElement('script')
     tag.src = script.src
     tag.id = script.bokeh_id
     tag.setAttribute('data-bokeh-model-id', script['data-bokeh-model-id'])
     tag.setAttribute('data-bokeh-doc-id', script['data-bokeh-doc-id'])
+    tag.onload = (event: Event) => {
+      // this._plot_ref = Bokeh.index[this._bokeh_id].model.document
+    }
 
     // wrap bokeh elements in div to apply css selector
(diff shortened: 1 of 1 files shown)


src/widget.ts
```diff
@@ -18,20 +18,23 @@ class DistributedUIElement extends Widget {
...
+    // store bokeh model id as private attr for access in onResize eventing
+    this._bokeh_id = script["data-bokeh-model-id"]
+
     let tag = document.createElement('script')
     tag.src = script.src
     tag.id = script.bokeh_id
     tag.setAttribute('data-bokeh-model-id', script['data-bokeh-model-id'])
     tag.setAttribute('data-bokeh-doc-id', script['data-bokeh-doc-id'])
+    tag.onload = (event: Event) => {
+      // this._plot_ref = Bokeh.index[this._bokeh_id].model.document
```
canavandl: this needs to set an attr on the parent

canavandl:
screen shot 2016-09-27 at 1 14 39 pm

mrocklin:
Thoughts on vertical resizing?

canavandl:
There was an issue where loading a new Bokeh element via script tag would override the `Bokeh.index`, so opening a second panel would cause the ref to the first plot to be lost.

The resizing now works because of a length onload callback on the script tag. Essentially, it has to wait long enough for bokehjs to load, the plot to render and be added to the global Bokeh.index, so that a reference the model can be stored as an attr on our Widget subclass.

The consequence of this is that the first plot won't resize (because the onResize event fires too quickly), but every subsequent new plot/panel drag event will make everything work as expected.

canavandl:
> Thoughts on vertical resizing?

I think the default sizing mode to element is `scale_width` and we might just have to change it to `scale_both`. I'll investigate

mrocklin:
I anticipate scale_both to be better for dask.distributed plots

canavandl:
superceded by #3

PR #3, opened 2016-09-28
Canavandl/individual routes

canavandl:
plot elements currently resize properly

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -15,23 +15,15 @@ const PORT = '8787'
 
 const SCRIPTS = [
   {
-    src: `http://${URL}:${PORT}/resource_profile/autoload.js?bokeh-autoload-element=0938e7ff-da78-4769-bf7f-b31d99fd9680`,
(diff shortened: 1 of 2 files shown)


canavandl:
todo (potentially in another PR):
* the Bokeh elements could use some padding inside the Phosphor containers, but simply adding css padding/margin interfered with the resizing
* a nicer menu
* better handling of selecting an already active plot from the command palette (currently it moves the plot to the 'main' area). It's not currently horrible, but could be better.

mrocklin:
@canavandl can we merge this?  I recommend that you take charge on pressing the green button on most of these things.

canavandl:
I think I need to make some changes to the distributed branch and get that merged first. Then this will be able to work with master and/or a distributed release.

canavandl:
The Bokeh team also resolved a memory leak in master that was causing issues with this extension: https://github.com/bokeh/bokeh/pull/5261

canavandl:
Will have to update this PR to add placeholder argument when https://github.com/bokeh/bokeh/pull/5295 is merged

canavandl:
superceded by #7

PR #11, opened 2016-11-08
Update version to 0.1.0

mrocklin:
Also require jupyterlab 0.8.0

@blink1073 any other versions that need to be changed here?


blink1073:
Looks good to me.

PR #18, opened 2018-08-29
Mega update

ian-r-rose:
Mostly complete update of the dask JupyterLab extension. Depends on dask/distributed#2185.

* Puts a dask dashboard launcher in the left panel, which can be used to launch different dashboard panels into the main area:
image
* Allows the user to set the URL of the dashboard bokeh server.
* Remembers the current URL, so it is persistent across page refreshes.
* Allows a default setting for the url via the JupyterLab settings system.

TODO: It would be nice to have a button to auto-detect the client dashboard url from the current notebook.

cc @mrocklin @canavandl


blink1073:
Nice!

mrocklin:
What is the right way to build and install a development lab-extension like this?  

The installation instructions we have in our README appear to be out-of-date.

After googling around I found https://jupyterlab.readthedocs.io/en/stable/developer/extension_dev.html#extension-authoring

But got this:

```
mrocklin@carbon:~/workspace/dask-labextension$ jupyter labextension install  # install the current directory as an extension
> /home/mrocklin/Software/node-v7.7.3-linux-x64/bin/npm pack /home/mrocklin/workspace/dask-labextension
jupyterlab-dask-0.1.0.tgz

Errored, use --debug for full output:
ValueError: "/home/mrocklin/workspace/dask-labextension" is not a valid extension:
schemaDir is empty: "schema"
```

blink1073:
Ah, it looks like the schema part of the `package.json` metadata needs to be removed since there are currently no schema.

mrocklin:
Yup.  That works.  Thanks @blink1073 .

I updated our dev installation instructions in the README.  If someone has a minute to take a look at that file that would be welcome.  I'm not sure if the installation requirements remain correct.  I also don't know what the minimum JLab version should be, although it looks like error reporting is good enough when installing the extension that we can possibly omit this from our instructions.

blink1073:
I'd say remove the version specifier.  You also don't need to enable an extension.  Otherwise, LGTM.

mrocklin:
OK, some questions:

So if I start a scheduler, open up a pane pointing to a plot, then close that scheduler and open another on the same port the plot in JLab becomes non-responsive.  This isn't too surprising, knowing when to refresh the iframe would be difficult here.  Given this, what is the best user experience that we can easily provide here?  My first attempt was to click on the plot button again, but that just highlighted the exisitng non-responsive plot again.  I found that I had to close the pane, then open it back up again with the plot-button.  

Entering `localhost:8787` into the `DASK DASHBOARD URL` field presents a message about no dashboard being present there.  The user needs to explicitly enter in the `http://`.  Is this something that we can add in for them if they are lazy, or is it better to be explicit?

Seeing the Dask icon present in the JupyterLab left-bar makes me unreasonably happy.  I'm inclined to color the buttons to a Dask orange as well.  Is here the right place to accomplish that?  (I'm still figuring out how to update things efficiently, and so haven't been able to try this out directly yet.)

```diff
diff --git a/style/index.css b/style/index.css
index 1d9cf90..a674d73 100644
--- a/style/index.css
+++ b/style/index.css
@@ -25,6 +25,8 @@
 .dask-DashboardListing-item button {
   flex: 1 1 auto;
   height: var(--dask-launch-button-height);
+  background-color: #ECB172;
+  border: 1px solid #ECB172;
 }
 
 .dask-DashboardListing-item button:disabled {
```

ian-r-rose:
The `schema` directory was something I forgot to add to the repo, sorry. It should be fixed now.

ian-r-rose:
With regards to detecting a shut-down cluster: I agree that the current behavior is not ideal. One possible way forward: I have implemented this somewhat (extremely?) hacky method for detecting whether the given URL is actually a bokeh server: https://github.com/ian-r-rose/dask-labextension/blob/6844c24574e5c44bbc0a53dc3e05bfbdbe6d63fb/src/widget.tsx#L344-L364 .
We *could* poll for an active bokeh server every few seconds to try to detect a shutdown. If that happens, we can show an error message and close the active plots.

My initial thinking was that it's better to be explicit about `http://`. It could be that the bokeh server is behind `https`, for instance. If there is a common place for it to be hosted, people can set that in the `"defaultURL"` setting to avoid having to type it all the time. Willing to be convinced otherwise, however. 

I am also a bit worried about loading `http` sites in iframes. It is likely that, as soon as JLab is served over `https`, we would get [mixed-content](https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content) warnings and errors. Eventually, I think the solution is to move to having a server extension so we can proxy the content server-side. @canavandl, is that something that has come up when dealing with bokeh server plots before?

With regards to button colors, yes, I see no problem making them dask orange. I would just check to make sure that it looks okay in both dark and light themes. The diff you have shown looks good to me.

ian-r-rose:
Some musing on setting the URL. It would be nice if there were a button that auto-detected the bokeh server URL for the given notebook. I was playing around with this, and I think sending something like this to the kernel might do the trick:
```python
try:
    from dask.distributed import default_client
    display(default_client.cluster.dashboard_link);
except:
    pass
```
Of course, this could affect the state of the kernel by making some imports and some having some extra iopub activity, so it makes me a bit nervous. I could mangle the name of the function (`import default_client as _private_dashboard_default_client_1234`) to cut down on the possibility of collision, but it would still be strictly possible.

On the other hand, we can't even guarantee that *completion* introspection has no side effects, so maybe I am overthinking it.

mrocklin:
We could also wrap the import it in a function so as to keep the scope small.

mrocklin:
Doesn't matter so much though

ian-r-rose:
Okay, I've added polling to check for dead connections, along with an error message for when that happens.

mrocklin:
Forgot the notebook link: https://gist.github.com/680dedaf04aa600791f8f71669a09ee3

ian-r-rose:
Yeah, I think that dropping the error message is probably okay. The visual of greyed out buttons may be enough to communicate a non-live dashboard. Also experimenting with the auto-detection right now...

mrocklin:
FWIW client.dashboard_link may not be a robust choice going forwards.
There are deployment scenarios where the client will have to query for this
information more directly.  I'm somewhat inclined to wait on this until
after some deployment questions have been ironed out in the various
dask-deployment communities (hpc/yarn/kubernetes).

mrocklin:
Happy either way though.  I suspect that the work you're doing now would
transfer easily.

ian-r-rose:
Do you anticipate that there will be *some* way to query for the dashboard link? Updating the python code snippet would be fairly straightforward. Otherwise I can drop that for now and declare this close-to-ready.

src/widget.tsx
```diff
@@ -0,0 +1,424 @@
...
+      this._input.value = newValue;
+      this._urlChanged.emit({ isValid: result, oldValue, newValue });
+      this._input.blur();
+      this.update();
+      if (!result) {
+        showErrorMessage(
+          'Invalid Dask URL',
+          Error(
+            `${newValue ||
+              'This'} does not appear to host a valid Dask dashboard`
```
mrocklin: This still triggers when you open JLab and you have panes open from your last session but without any scheduler running yet.

mrocklin:
> Do you anticipate that there will be some way to query for the dashboard link? Updating the python code snippet would be fairly straightforward. Otherwise I can drop that for now and declare this close-to-ready.

Yes, that's a fair assumption.

ian-r-rose:
I have added a search button that tries to query the currently active kernel for the dask dashboard URL. If none is found, it does nothing:
image
If we find it is unreliable, it's easy enough to remove.

mrocklin:
Functionally this works quite well for me.  My coloring is a bit off though:

image

ian-r-rose:
What is the problem with the coloring? I used the HTML codes you posted above, and black text for the wording.

ian-r-rose:
Ah, I see: the search icon. Fix incoming.

mrocklin:
Right, you beat me to the response :)

mrocklin:
This works quite well for me.  I can't think of anything I would want to change.

The check for the file `dask_horizontal.svg` will change in the next release.  I'll push up a change here once https://github.com/dask/distributed/pull/2223 is merged in.

ian-r-rose:
Great. Eventually that check should probably be replaced with a server-side component to this extension to handle the proxy-ing. That would also help to cut down on mixed-content issues in HTTP/HTTPS settings. For now this seems to work okay.

src/widget.tsx
```diff
@@ -397,7 +397,8 @@ export namespace DaskDashboardLauncher {
 
   export const DEFAULT_ITEMS = [
     { route: 'individual-graph', label: 'Graph' },
-    { route: 'individual-load', label: 'Load' },
+    { route: 'individual-nbytes', label: 'Memory Use' },
+    { route: 'individual-nprocessing', label: 'Processing Tasks' },
```
mrocklin: @ian-r-rose the "Load" plot was split into two upstream.  I found that I needed to update these manually.  I know that at one point we discussed getting this information from Dask itself.  I'm inclined to leave this as future work, but wanted to point it out here.

mrocklin:
From my perspective this is good.  Upstream has stabilized.  I'm inclined to merge this and release.  Any objections or additional work that should be done before that happens?

src/widget.tsx
ian-r-rose: Yeah, I had wanted to do that, but was stymied by the CORS issues. Once we move some logic into a server-side proxy we should be able to serve the individual routes dynamically, but for now I think this is the only way.

ian-r-rose:
No objections, I think we are good to go.

ian-r-rose:
Shall I publish to npm?

mrocklin:
OK then, this is in. Thank you @ian-r-rose for doing this work.  You should have an invitation giving you commit rights to this repository.

> Shall I publish to npm?

Yes, that would be most welcome.  Presumably also to PyPI?  I don't understand the release process of JupyterLab plugins well, so I'd be happy to defer to you on this.

I can work on some docs within the mainline Dask docs.

ian-r-rose:
Since there is no python package, there is nothing to publish to PyPI. The user will install the extension using `jupyter labextension install dask-labextension`. Speaking of which, should we stick with that name?

mrocklin:
What are normal conventions behind JLab extension names?  I'm fine with the
name, but it probably makes less sense given that `labextension` is already
in the command to install it

ian-r-rose:
I would say that strong naming conventions have not been established. I have personally been naming things of the form `jupyterlab-dask`, but that is not universal. Ultimately not a huge deal either way.

mrocklin:
That would be fine with me.  So too would `dask-jupyterlab` (`dask-foo` is the convention behind dask sub-projects).  I'm inclined to leave this decision with you as primary author if you're up for it.

ian-r-rose:
I kept it as-is. So the install would be `jupyter labextension install dask-labextension`. It is now live!

mrocklin:
dashboard-jupyterlab

Uploading an image here for posterity

Issue #21, opened 2018-09-01
Fresh install, can't figure out how to view


stsievert:
I've installed this extension with the extension manager in Jupyterlab 0.34.7.

<img width="233" alt="screen shot 2018-09-01 at 2 43 54 pm" src="https://user-images.githubusercontent.com/1320475/44949369-6d8d5c00-ae1f-11e8-9625-eedd20df598a.png">

When I try to view the panes dask-labextension presents, I see greyed out buttons:

<img width="522" alt="screen shot 2018-09-01 at 2 46 46 pm" src="https://user-images.githubusercontent.com/1320475/44949397-df65a580-ae1f-11e8-874e-1ed035546472.png">

When I try to drag them or double click on them, nothing happens. I've tried changing "Dask Dashboard URL" to `http://127.0.0.1/status`, the value presented by `client`.

How do I view the panes dask-labextension presents?

stsievert:
I tried that, but no luck. When I start up JupyterLab, I get a message

``` shell
# right below "Copy/paste this URL into your browser when you connect ..."
[W 18:56:10.034 LabApp] 404 GET /lab/api/settings/jupyterlab-dask:plugin?1535846169700 (::1): Schema not found: '/Users/scott/anaconda3/share/jupyter/lab/schemas/jupyterlab-dask/plugin.json'
[W 18:56:10.035 LabApp] Schema not found: '/Users/scott/anaconda3/share/jupyter/lab/schemas/jupyterlab-dask/plugin.json'
[W 18:56:10.037 LabApp] 404 GET /lab/api/settings/jupyterlab-dask:plugin?1535846169700 (::1) 4.93ms referer=None
```

When I visit `/Users/scott/anaconda3/share/jupyter/lab/schemas/`, I see a directory named `dask-labextension`, not `jupyterlab-dask`. `dask-labextension` contains `plugin.json`

It looks like Jupyter is looking for a directory `jupyterlab-dask`, not `dask-labextension`. I tried copy/pasting `dask-labextension` to `jupyterlab-dask`. That got rid of the error message above, but didn't resolve the issue (the buttons are still greyed out).

dhirschfeld:
I also can't get this extension to work.

I also get a 404 on `jupyterlab-dask`
```
/user/dhirschfeld/lab/api/settings/jupyterlab-dask:plugin?1535956133900 404 (Not Found)
```

When I enter the dask dashboard url it then says it's not valid:
```
[...]

PR #22, opened 2018-09-04
Usability Fixes

ian-r-rose:
Fixes https://github.com/dask/dask-labextension/issues/21#issuecomment-418014155
Fixes https://github.com/dask/dask-labextension/issues/21#issuecomment-418014155

This should be published as a patch-release.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,7 +14,7 @@ import { INotebookTracker, NotebookPanel } from '@jupyterlab/notebook';
 
 import { Kernel, KernelMessage } from '@jupyterlab/services';
 
-import { DaskDashboardLauncher } from './widget';
+import { DaskDashboardLauncher, normalizeDashboardUrl } from './widget';
 
 import '../style/index.css';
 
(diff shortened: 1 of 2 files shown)


dhirschfeld:
I gave this branch a go and I think it fixed the incorrect naming but I'm still getting CORS errors so will give `nbserverproxy` a go...

ian-r-rose:
Yes, I didn't attempt to fix any CORS issues with this PR. You *may* be able to disable CORS policies on a per-site basis for testing purposes, but that is probably not a good long-term solution.

ian-r-rose:
It's looking like we should set up some best-practices around proxying the dashboard URL to be within the notebook server origin sooner rather than later. Do you know who has relevant experience doing that?

mrocklin:
The Pangeo community is probably the best bet.  Typically we've just
install nbserverproxy in our docker images and then used that.  This hasn't
been particularly sophisticated so far.

ian-r-rose:
Published as 0.1.1

mrocklin:
Thanks @ian-r-rose

PR #25, opened 2018-09-08
This was mistakenly uninitialized.

ian-r-rose:
This should fix some of the layout restoration issues seen in dask/dask-examples#38


ian-r-rose:
Published as v0.1.2

Issue #35, opened 2018-12-21
Automatically create client to current cluster in the background


mrocklin:
I'm interested in a workflow where users never have to create a Dask client, it just gets created for them silently in the background based on the current active Cluster in the server extension.  So an order of operations might look like the following:

1.  Open up JupyterLab
2.  Open up Notebook and ipython kernel
3.  Create cluster using the dask-labextension pane
4.  When we create that cluster it automatically creates a client in the notebook kernel without us doing anything.  All dask.compute calls automatically use that cluster/client
5.  Create a new cluster, have it automatically inject code to change the default client in the kernel
6.  Start a new kernel, either by creating a new notebook or by restarting your kernel
7.  The extension sees this and automatically injects code to create a new client

As a result, the user always uses the cluster specified in the labextension.  They have to explicitly choose *not* to use this if they want to use the local scheduler or a client of their own making.

This behavior would be off by default, but we could turn it on for things like tutorials, pangeo, and so on.

Is this useful, or is it too magical?

cc @ian-r-rose @jhamman @rabernat

ian-r-rose:
Fixed by #36

PR #36, opened 2018-12-22
Greedy cluster client

ian-r-rose:
Work towards having all active notebooks and consoles get references to a client for the currently active Dask cluster. The current code injected into a kernel is 
```python
import dask; from dask.distributed import Client
dask.config.set({'scheduler-address': '${model.scheduler_address}'})
client = Client()
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@jupyterlab/codeeditor": "^0.19.1",
     "@jupyterlab/console": "^0.19.1",
     "@jupyterlab/coreutils": "^2.2.1",
+    "@jupyterlab/mainmenu": "^0.8.1",
     "@jupyterlab/notebook": "^0.19.1",
     "@jupyterlab/services": "^3.2.1",
     "@phosphor/coreutils": "^1.3.0",
--- a/schema/plugin.json
+++ b/schema/plugin.json
@@ -8,6 +8,12 @@
       "type": "string",
       "title": "Default URL for the Dask Dashboard Webserver",
       "default": ""
+    },
+    "greedyClusterClient": {
+      "type": "boolean",
+      "title": "Greedy Cluster Client",
+      "description": "If set to true, every notebook and console will automatically have a dask client for the active cluster injected into the kernel under the name 'client'",
+      "default": false
     }
   },
   "type": "object"
--- a/schema/plugin.json
+++ b/schema/plugin.json
@@ -9,9 +9,9 @@
       "title": "Default URL for the Dask Dashboard Webserver",
       "default": ""
     },
-    "greedyClusterClient": {
+    "autoStartClient": {
       "type": "boolean",
-      "title": "Greedy Cluster Client",
+      "title": "Auto-Start Client",
       "description": "If set to true, every notebook and console will automatically have a dask client for the active cluster injected into the kernel under the name 'client'",
       "default": false
     }
--- a/src/clusters.tsx
+++ b/src/clusters.tsx
@@ -112,6 +112,13 @@ export class DaskClusterManager extends Widget {
   }
 
   /**
+   * Set an active cluster by id.
+   */
+  setActiveCluster(id: string): void {
+    this._setActiveById(id);
+  }
+
+  /**
    * A signal that is emitted when an active cluster changes.
    */
   get activeClusterChanged(): ISignal<
@@ -126,6 +133,13 @@ export class DaskClusterManager extends Widget {
    */
   get clusters(): IClusterModel[] {
     return this._clusters;
+  }
+
+  /**
+   * Refresh the current list of clusters.
+   */
+  async refresh(): Promise<void> {
+    await this._updateClusterList();
   }
 
   /**
@@ -220,6 +234,17 @@ export class DaskClusterManager extends Widget {
     );
     const data = (await response.json()) as IClusterModel[];
     this._clusters = data;
+
+    // Check to see if the active cluster still exits.
+    // If it doesn't, or if there is no active cluster,
+    // select the first one.
+    const active = this._clusters.find(
+      c => c.id === (this._activeCluster && this._activeCluster.id)
+    );
+    if (!active) {
+      const id = (this._clusters[0] && this._clusters[0].id) || '';
+      this._setActiveById(id);
+    }
     this.update();
   }
 
--- a/src/clusters.tsx
+++ b/src/clusters.tsx
@@ -1,10 +1,14 @@
 import { Toolbar, ToolbarButton } from '@jupyterlab/apputils';
 
+import { IChangedArgs } from '@jupyterlab/coreutils';
+
 import { ServerConnection } from '@jupyterlab/services';
 
 import { JSONObject, JSONExt } from '@phosphor/coreutils';
 
 import { Message } from '@phosphor/messaging';
+
+import { ISignal, Signal } from '@phosphor/signaling';
 
 import { Widget, PanelLayout } from '@phosphor/widgets';
 
@@ -36,7 +40,17 @@ export class DaskClusterManager extends Widget {
         return;
       }
       options.setDashboardUrl(`dask/dashboard/${cluster.id}`);
-      this._activeClusterId = id;
+
+      const old = this._activeCluster;
+      if (old && old.id === cluster.id) {
+        return;
+      }
+      this._activeCluster = cluster;
+      this._activeClusterChanged.emit({
+        name: 'cluster',
+        oldValue: old,
+        newValue: cluster
+      });
       this.update();
     };
 
@@ -91,7 +105,24 @@ export class DaskClusterManager extends Widget {
   }
 
   /**
-   * Get the currently active clusters known to the manager.
+   * The currently selected cluster, or undefined if there is none.
+   */
+  get activeCluster(): IClusterModel | undefined {
+    return this._activeCluster;
+  }
+
+  /**
+   * A signal that is emitted when an active cluster changes.
+   */
+  get activeClusterChanged(): ISignal<
+    this,
+    IChangedArgs<IClusterModel | undefined>
+  > {
+    return this._activeClusterChanged;
+  }
+
+  /**
+   * Get the current clusters known to the manager.
    */
   get clusters(): IClusterModel[] {
     return this._clusters;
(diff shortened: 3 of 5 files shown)


ian-r-rose:
Fixes #35

mrocklin:
For context.  @ian-r-rose and I were talking about ways to enable administrators to create clusters for users that they never have to interact with.  

For example I think that the combination of this PR along with #37 will allow for Pangeo or example.dask.org users to just start with `import xarray` or `import iris` and not think about starting Dask clusters or Dask clients at all.  Hopefully we can hide that boilerplate from them.

cc @jacobtomlinson @yuvipanda @jhamman

mrocklin:
To be more explicit

image

ian-r-rose:
I agree that this behavior should not be on by default. I think the best way to approach it would be via the JupyterLab settings system.

jhamman:
@ian-r-rose - this looks really cool. I'll try to give it a test spin next week. Also pinging @jacobtomlinson and @niallrobinson for comment since this is something they have been interested in.

ian-r-rose:
@jhamman Did you get a chance to play with this?

mrocklin:
I suggest the name auto-start rather than greedy.  

Is there a way for us to control this in a Jupyter config file?  If so, what does that config file look like?

mrocklin:
Also, we can maybe drop the term `Cluster` from the name?  Maybe add Dask (if things aren't already namespaced)

ian-r-rose:
Thanks for the suggestions @mrocklin. If I understand correctly, you are suggesting something like
1. Name the setting `autoStartClient`
2. Label the menu item `Auto-Start Dask Client`
Is that right?

This is driven by the client-side, so it can't be controlled in the server-side `juptyer_notebook_config.py`. That setting, however, is stored as a JSON file on disk, so it can be seeded into a docker image if needed.

ian-r-rose:
Sure, that makes sense to me. The setting name is a bit more targeted towards experts and admins, so I might advocate being a more specific there and retaining "Client." The menu item is much more user-facing, so removing "Client" works for me.

Regarding including "dask": the setting name is already namespaced by being in the Dask settings,  but the menu label should have "Dask" in it, as you suggest.

mrocklin:
OK.  That makes sense to me.

rabernat:
Ping @jbusecke for a review of this. Julius is a postdoc at Princeton and a heavy user of pangeo on HPC. He mentioned that he was interested in this extension. Perhaps he can take this PR for a spin and give some feedback.

jbusecke:
Oh this looks super sweet! I will definitely try to take this for a spin on the HPCs I am working on in the next days!

jhamman:
I setup a test binder with this branch today: [Binder](https://binder.pangeo.io/v2/gh/jhamman/pangeo-example-notebooks/test_dask_extension)

Question: it sounds like this feature is meant to be off by default. How to I set it to on?

ian-r-rose:
Hi @jhamman, you can toggle it in the Settings menu under "Auto-Start Dask".

ian-r-rose:
Although I am having a hard time with that binder link because clusters are not starting for some reason.

jhamman:
The logs I'm getting on the binder is: 
```
[W 04:12:24.249 LabApp] object KubeCluster can't be used in 'await' expression
[I 04:12:25.435 LabApp] Client sent subprotocols: ['']
[I 04:12:25.436 LabApp] Trying to establish websocket connection to ws://127.0.0.1:8787/individual-progress/ws?bokeh-protocol-version=1.0&bokeh-session-id=CVEoIFIvycC2DEGuFlS9jd9MsdPjBoirUFNwSAjkrT6n
[I 04:12:25.440 LabApp] Websocket connection established to ws://127.0.0.1:8787/individual-progress/ws?bokeh-protocol-version=1.0&bokeh-session-id=CVEoIFIvycC2DEGuFlS9jd9MsdPjBoirUFNwSAjkrT6n
[I 04:12:25.725 LabApp] Client sent subprotocols: ['']
[I 04:12:25.726 LabApp] Trying to establish websocket connection to ws://127.0.0.1:8787/individual-task-stream/ws?bokeh-protocol-version=1.0&bokeh-session-id=Ku90IbfLbhki6h46zNHhJCXdTA1Ge0Jvjzc1uVUSDxgL
[I 04:12:25.729 LabApp] Websocket connection established to ws://127.0.0.1:8787/individual-task-stream/ws?bokeh-protocol-version=1.0&bokeh-session-id=Ku90IbfLbhki6h46zNHhJCXdTA1Ge0Jvjzc1uVUSDxgL
```
should I be working off development versions of dask/distributed/dask-kubernetes?

mrocklin:
OK, it looks like we'll have to make dask-kubernetes more async friendly.
This is probably on me (or maybe @yuvipanda depending on when he gets to
some async kubernetes work)

mrocklin:
https://github.com/dask/dask-kubernetes/pull/116 may work?  Untested

mrocklin:
You'll also need distributed master

ian-r-rose:
@jhamman Yes, you can set it from outside of lab. The settings file is a JSON file on disk, so you can seed the docker image with that in place (I don't think the `start` script should even be necessary). I'll take a crack at it in your test branch today sometime.

ian-r-rose:
[here](https://github.com/ian-r-rose/pangeo-example-notebooks/commit/00af30a904f529b31442a2fc152dec4af460ca04) is a commit that auto-sets the setting in your binder example. It seems to work well, but there is still something broken with the cluster config...

mrocklin:
Just to update this PR I think that we should get async-kubernetes working in dask-kubernetes, and then revisit this.  

It would also be good to test this with dask-jobqueue and dask-yarn.  Mostly this makes the request that clusters can be started and stopped asynchronously.  I suspect that this is already true for Dask-Jobqueue, but it would be good to test in practice (cc @guillaumeeb).  I don't know the current status in dask-yarn (cc @jcrist )

ian-r-rose:
No rush on this, happy to let people kick the wheels.

That being said, isn't the async cluster starting already in master (#37)?

mrocklin:
Hrm, indeed...

guillaumeeb:
Hi,

Not sure what is meant by
> Mostly this makes the request that clusters can be started and stopped asynchronously

Unfortunately I did not have the time to test dask-labextension at all yet. I'm a little behind on this, we're still using standard Notebook with our Jupyterhub at CNES (due to a probably not really complicated problem with JLab I need to work on). 

This looks interesting though, I'm often launching several clusters from several Notebooks, so if I could use only one, and automatically, that would be great! I'm putting this on my todo-list, but don"t wait on me to merge...

ian-r-rose:
Hi @guillaumeeb, thanks for sharing your use-case! I'd like to hear how this feature works for you when you get the chance to try it.

> 
> Not sure what is meant by
> 
> > Mostly this makes the request that clusters can be started and stopped asynchronously
> 

This extension expects that different cluster implentations (e.g. `LocalCluster` or `KubeCluster`) can be started without blocking the main event loop. This means, for instance, that JupyterLab startup time won't be delayed by setting up any initial clusters: they can finish in their own time and be populated when they are ready.

We got a little bit ahead of ourselves by expecting that, however, since there is not a critical mass of implementers who have made cluster startup async yet. So now there is a bit of catch-up going on to make sure that works.

mrocklin:
Note that I got dask-kubernetes mostly working asynchronously if anyone wants to give it a shot

    pip install git+https://github.com/dask/dask-kubernetes@dev

mrocklin:
Where by "I" I actually mean "Yuvi and I"

guillaumeeb:
So I finally managed to deploy extensions into Jupyterlab behind my corporate proxy. I'm just beginning testing dask-labextension. Before I try this PR, I just wanted to know if my envrionment was working as intended. Typical steps I take right now are:

1. Start a cluster inside my notebook
1. Indicate scheduler URL in dask-labextension left panel
1. Open views on my lab environment

Using Pangeo on binder, the notebook was started with an already defined layout: how do I do this?
I was under the impression that once a Cluster was started, dask-labextension views were automatically connected to them, isn't that true? Is the second step above always needed?

ian-r-rose:
Hi @guillaumeeb, this functionality is intended to work with clusters managed by the extension (rather than ones launched in your notebook). It currently works with `LocalCluster`, and (possibly) `KubeCluster`. Basically, the cluster implementation needs to satisfy the `Cluster` interface, and be possible to start asynchronously (i.e. `cluster = await MyCluster(*args)`). We got a bit ahead of ourselves in requiring the async startup here, so it's possible that your deployment doesn't yet fit that usage.

If those preconditions are met, then all new notebooks should get a client for that cluster auto-injected into the current python kernel session (when the option turned on).

guillaumeeb:
>  and be possible to start asynchronously (i.e. `cluster = await MyCluster(*args)`)

Okay, that helps a lot to know what is needed here.

> and (possibly) `KubeCluster`

Why possibly? I personnaly maintain dask-jobqueue, which provides implementations of Cluster, not sure about the asynchronous part. Are these the only two requirements? How do we launch a cluster with the extension? I guess I should use the **New** button, but nothing happens when I try to click on it on my setup.

Maybe I should rather open another issue to discuss the questions from my previous comment on standard dask-labextension behavior, not really related to this PR?

ian-r-rose:
> > and be possible to start asynchronously (i.e. `cluster = await MyCluster(*args)`)
> 
> Okay, that helps a lot to know what is needed here.
> 
> > and (possibly) `KubeCluster`
> 
> Why possibly? 

I say possibly because @mrocklin and @yuvipanda just put a bunch of work into updating it, but I don't think it has been tested in this context yet.

> I personnaly maintain dask-jobqueue, which provides implementations of Cluster, not sure about the asynchronous part. Are these the only two requirements? How do we launch a cluster with the extension? I guess I should use the **New** button, but nothing happens when I try to click on it on my setup.

The function to launch a new cluster is defined here: https://github.com/dask/dask-labextension/blob/910f5fff9e4d50f79a8c42e0ecd14464bbc2a177/dask_labextension/manager.py#L23-L41

The intention is that those are the only two requirements (implements `Cluster` and is `await`able), but we likely have some kinks to work out to make it widely usable. Are there any errors in the notebook logs when you click the "New" button?

> 
> Maybe I should rather open another issue to discuss the questions from my previous comment on standard dask-labextension behavior, not really related to this PR?

That would be great, thanks @guillaumeeb.

mrocklin:
This is in.  My apologies for the long delay @ian-r-rose !

michaelaye:
It is unclear to me where I get the seemingly hashed Dask Dashboard URL from that is visible in above screenshots? I tried to simply set it to http://127.0.0.1:8787/status, after which all the buttons lit up, and clicking on them opened a new jlab view, but no content was displayed. I guess I'm missing something? My default browser display of the status page works fine.

PR #37, opened 2018-12-22
Add initial cluster parameters and instances

mrocklin:
This is joint work with @ian-r-rose 

This adds a few things:

-  The ability within the model to create workers and set adaptivity when a cluster is created
-  Configuration to trigger this by default
-  The ability to start clusters on startup
-  Additional asynchronous behavior on the server side

It currently depends on two PRs:

-  https://github.com/dask/distributed/pull/2437
-  https://github.com/dask/dask/pull/4324

To start reviewing I recommend looking at the changes to the default labextension.yaml file and then the tests (we have tests now!)

### DIFF
--- a/dask_labextension/clusterhandler.py
+++ b/dask_labextension/clusterhandler.py
@@ -17,12 +17,12 @@ class DaskClusterHandler(APIHandler):
     """
 
     @web.authenticated
-    def delete(self, cluster_id: str) -> None:
+    async def delete(self, cluster_id: str) -> None:
         """
         Delete a cluster by id.
         """
         try:  # to delete the cluster.
-            val = manager.close_cluster(cluster_id)
+            val = await manager.close_cluster(cluster_id)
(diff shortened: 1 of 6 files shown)


mrocklin:
cc @jhamman @rabernat @jacobtomlinson @yuvipanda

mrocklin:
@ian-r-rose are there things that you would like to see here before we merge?

ian-r-rose:
I think this is in pretty good shape. If you are happy with it, I think we should merge and move forwards. I have a couple of other PRs to get in (#36 and one with a more friendly unconnected-dashboard-panel), and we can think about cutting another release.

mrocklin:
Sounds good to me.  If it's alright I'd like to leave the merging process
to you in case you'd prefer to preserve history.

ian-r-rose:
Sounds good. Onwards, thanks @mrocklin!

PR #38, opened 2018-12-29
Friendlier empty panel

ian-r-rose:
This adds an overlay to the dashboard plots when there is no dashboard found. It prevents some ugly 404 errors from being shown to the user (at least for the most part).

image

cc @jhamman

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -4,7 +4,7 @@ import { PageConfig, URLExt } from '@jupyterlab/coreutils';
 
 import { ServerConnection } from '@jupyterlab/services';
 
-import { JSONObject } from '@phosphor/coreutils';
+import { JSONExt, JSONObject } from '@phosphor/coreutils';
 
 import { Message } from '@phosphor/messaging';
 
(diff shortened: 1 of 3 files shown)


mrocklin:
This seems much nicer than a 404 :)

I anticipate new users will be concerned about the "Cannot find dashboard" message when they first start up and haven't yet started a cluster.

We might either have a message like "Start a Dask cluster to see dashboard" or maybe drop the text altogether (I have a slight preference for this).

ian-r-rose:
Sure, happy to go with either of those suggestions. I'll wait to see if @jhamman wants to weigh in, then update.

jhamman:
@ian-r-rose - thanks for this. I'd be happy with either of @mrocklin's suggestions.

ian-r-rose:
Okay, I've removed the text from the inactive panel. This is ready for a spin/review.

ian-r-rose:
Moving forwards with this one. It's mostly cosmetic, but a nice feature to avoid confusing users and exposing them to the iframe implementation details.

stsievert:
Does this also prevent the 404 messages that are presented on the terminal?

PR #39, opened 2018-12-30
Increase testing

ian-r-rose:
Getting familiar with some of the testing infrastructure that @mrocklin wrote in #37 by writing more tests and making sure that the public APIs are up to the task of being tested. A couple of notes from the experience:

1. I have added an `initialized` future to the cluster manager which can be awaited. This seems cleaner to me than polling for the `manager._clusters` dictionary.
2. The `adapt` and `scale` functions are not awaitable, which makes writing tests for them pretty tricky (as evidenced by the [tests](https://github.com/dask/distributed/blob/5655381ac570065ab1c742b88dae82f73b980ece/distributed/deploy/tests/test_local.py#L391-L435) in `distributed`). It would definitely be nice to be able to know when, and whether, the scaling operation was successful. This also touches on the discussions in dask/distributed#2235 and dask/distributed#2257.


dask_labextension/tests/test_manager.py
```diff
@@ -0,0 +1,115 @@
...
+            # assert not model.get('adapt')
+
+@gen_test()
+async def test_initial():
+    with dask.config.set({
+        'labextension.defaults.kwargs': {'processes': False},  # for speed
+        'labextension.initial': [{'name': 'foo'}],
+    }):
+        async with DaskClusterManager() as manager:
+            await manager.initialized
```
mrocklin: Thoughts on using just `await manager` here instead?  We do this with other Dask objects.  This would require us to implement an `__await__` method on manager that returned the initialized Future.

I would also be fine with calling `await self` in `__aenter__`

mrocklin:
> The adapt and scale functions are not awaitable, which makes writing tests for them pretty tricky (as evidenced by the tests in distributed). It would definitely be nice to be able to know when, and whether, the scaling operation was successful. 

Asking the Cluster subprojects to return a Future from scale seems reasonable to me.  

What would awaiting on adapt look like?  Is it "wait until there are no more new/old workers in flight and we seem to be at a steady state?"

ian-r-rose:
That's a good point about adapt. Maybe it's best to just think about scale for the time being, since that's where I ran into trouble.

dask_labextension/tests/test_manager.py
ian-r-rose: Sure, happy to convert this to more idiomatic code, thanks for the suggestions.

dask_labextension/tests/test_manager.py
```diff
@@ -0,0 +1,151 @@
...
+        # Test asynchronous starting of clusters outside of a context
+        manager = DaskClusterManager()
+        assert len(manager.list_clusters()) == 0
+        await manager
+        assert len(clusters) == 1
+        assert clusters[0]["name"] == 'foo'
+        await manager.close()
+
+        manager = await DaskClusterManager()
+        assert len(clusters) == 1
```
mrocklin: Should this be `manager.clusters`?

mrocklin:
Travis-CI has been turned on for pull requests to this repository

dask_labextension/tests/test_manager.py
ian-r-rose: Yep, good catch.

ian-r-rose:
:+1: thanks

mrocklin:
Tests pass, we're good to go.

dask_labextension/manager.py
```diff
@@ -23,8 +23,12 @@
 async def make_cluster(configuration: dict) -> Cluster:
     module = importlib.import_module(dask.config.get('labextension.factory.module'))
     Cluster = getattr(module, dask.config.get('labextension.factory.class'))
+
+    kwargs = dask.config.get('labextension.factory.kwargs').copy()
+    if 'n_workers' not in kwargs and utils.has_keyword(Cluster, 'n_workers'):
```
ian-r-rose: This seems to make an assumption about the keyword arguments of the factory function that we had not been making before, which was quite general. Do you ask that all cluster constructors implement `n_workers`? Might this be fragile if `n_workers` does something different than we expect?

ian-r-rose:
Sounds good, thanks for bringing this over the finish line.

mrocklin:
I'm inclined to merge.  Thoughts @ian-r-rose ?

Also, how would you like to merge in general?  Merge commits?  Rebase and merge?  Squash?

ian-r-rose:
I think this is ready,  merge away! I generally prefer to not squash for larger PRs, but for stuff of this size I don't have strong opinions.

PR #47, opened 2019-02-10
Drag drop cluster connection

### LINKED ISSUE #44: Code-cell drag & drop functionality for cluster
The greedy/auto cluster client is a little too black-magic for my tastes but I think what would be very cool is if you could drag a cluster into a notebook and have it create a new code cell:
```python
import dask
from dask.distributed import Client

client = Client('tcp://address:port')
```

ian-r-rose:
Fixes #44.

This adds drag/drop of the client connection code from the cluster listing in the side panel. It works with notebooks, code consoles, and text files. Example:
dask-drag-drop

We now provide *three* ways to inject this code: via a button, via a context menu item, and via drag/drop.


mrocklin:
That's pretty awesome.  cc @jhamman

jhamman:
I dig it. I expect this to be a widely used feature.

PR #52, opened 2019-03-29
Add build status, version, download and dependency badges

dbast:


ian-r-rose:
Looks good, thanks @dbast

PR #57, opened 2019-04-14
Fix initial cluster

### LINKED ISSUE #56: Configuration broken for kwargs that have underscores
We currently ask users to configure clusters startup using `kwargs` in a config file. Unfortunately, dask's config normalizes underscores to hyphens, making the keys invalid python identifiers..

We can deal with it as a special case here, and replace hyphens with underscores before passing in the kwargs, but a nicer solution might be to have this fixed upstream.

xref https://github.com/dask/dask/pull/4422, https://github.com/dask/dask/issues/4366

ian-r-rose:
Fixes #56

### DIFF
--- a/dask_labextension/config.py
+++ b/dask_labextension/config.py
@@ -10,6 +10,6 @@ fn = os.path.join(os.path.dirname(__file__), 'labextension.yaml')
 dask.config.ensure_file(source=fn)
 
 with open(fn) as f:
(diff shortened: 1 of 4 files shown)


mrocklin:
+1 from me.  OK to merge?

ian-r-rose:
Yep, I plan to make a quick patch release with this.

Issue #49, opened 2019-02-19
Switching to jupyter-server-proxy


jakirkham:
IIUC this is using `nbserverproxy` currently. As `jupyter-server-proxy` is the successor of `nbserverproxy` recently and made a 1.0 release, it would be good to switch to using `jupyter-server-proxy` instead.

ian-r-rose:
Yes, switching to `jupyter-server-proxy` is a priority. I would say that is currently blocked by jupyterhub/jupyter-server-proxy#70

ian-r-rose:
Fixed by #58.

PR #58, opened 2019-04-27
Replace internal fork of nbserverproxy with jupyter-server-proxy

yuvipanda:


yuvipanda:
The idea is that when a new cluster gets created, we register a new tornado handler returned by make_dashboard_handler. We can unregister it when cluster is deleted. This also lets us use cluster names rather than IDs if needed, and remove state that is matching cluster IDs with clusters.

yuvipanda:
@ian-r-rose I also haven't actually tested this :D But it should work, and I copy pasted this code from another project I just did...

ian-r-rose:
One thing I was at least trying to leave the door open to was dashboard servers that were running other places than `localhost`. I'm not aware of anywhere that this was actually done, but it seemed like it was a possibility for some remote cluster setups. Does this `LocalProxyHandler` approach remove that possibility?

yuvipanda:
Nope it should be fine. We can make that configurable as the next step.

mrocklin:
FYI @quasiben .  Nothing for you to do here, but wanted to make you aware
of the jupyter-server-proxy and how it's used in dask-labextension.

quasiben:
👍

ian-r-rose:
Fixes #49. cc @jakirkham

dhirschfeld:
NB: My current distributed setup has a remote scheduler/dashboard. I am in the process of moving to k8s though but IIUC there are plans to allows remote schedulers there too

ian-r-rose:
Thanks for the info @dhirschfeld. @yuvipanda, sounds like we should move remote-server proxying towards the front-burner.

dhirschfeld:
No need to rush on my account - it's just something to keep in mind.

***xref:*** https://github.com/dask/dask-kubernetes/issues/84

ian-r-rose:
Okay, this should be good to go from my end.

@dhirschfeld I think this should continue working with your remote dashboard setup, but let us know if there are any troubles.

ian-r-rose:
Moving forwards with this. Thanks @yuvipanda!

jakirkham:
Thanks so much all for working on this! 😄

PR #59, opened 2019-05-09
Better failures

ian-r-rose:
This is an attempt to show more useful error messages when things go wrong. In particular, it tries to give error dialogs when clusters fail to start, stop, or scale, and it suggests that the server extension may not be installed correctly when not found.


mrocklin:
Woot

PR #65, opened 2019-06-16
If the user has manually set the proxy URL for the dashboard link, don't proxy again.

ian-r-rose:
Users frequently set `dashboard.link` in their dask config to that of the default `jupyter_server_proxy` URL. In that case, double proxying seems to make things not function properly.

This handles that case by checking if the dashboard link is *already* proxying, and simply using it if that is the case.


ian-r-rose:
Plan to merge and publish a quick patch release in the interests of getting the `dask-examples` repo back online ASAP.

PR #73, opened 2019-07-28
Use cluster.scheduler_info rather than cluster.scheduler

mrocklin:
Supercedes https://github.com/dask/dask-labextension/pull/72

This depends on https://github.com/dask/distributed/pull/2902 , which
adds a `Cluster.scheduler_info` attribute to clusters which holds
necessary scheduler information.  We prefer this over querying a
Scheduler object directly in case that scheduler is not local, as in
increasingly becoming the case.

### DIFF
--- a/dask_labextension/manager.py
+++ b/dask_labextension/manager.py
@@ -268,6 +268,10 @@ def make_cluster_model(
         info = cluster.scheduler_info
     except AttributeError:
         info = cluster.scheduler.identity()
+    try:
+        cores = sum(d["nthreads"] for d in info["workers"].values())
+    except KeyError:  # dask.__version__ < 2.0
+        cores = sum(d["ncores"] for d in info["workers"].values())
     assert isinstance(info, dict)
     model = dict(
         id=cluster_id,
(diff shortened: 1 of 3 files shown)


mrocklin:
OK, I think that this is good to go.  It should be safe both for old and new releases.

dask_labextension/manager.py
```diff
@@ -264,16 +264,21 @@ def make_cluster_model(
...
         scheduler_address=cluster.scheduler_address,
         dashboard_link=cluster.dashboard_link or "",
-        workers=len(cluster.scheduler.workers),
+        workers=len(info["workers"]),
         memory=utils.format_bytes(
-            sum(ws.memory_limit for ws in cluster.scheduler.workers.values())
+            sum(d["memory_limit"] for d in info["workers"].values())
         ),
-        cores=sum(ws.ncores for ws in cluster.scheduler.workers.values()),
+        cores=sum(d["nthreads"] for d in info["workers"].values()),
```
ian-r-rose: What is the intended compatibility profile for this? I am getting failures here due to `nthreads` not being available in my workers with dask v1.2 and distributed v1.28.

mrocklin: 2.0+, but it's fairly easy to go handle this to go back further.  I'll push up a small fix.

ian-r-rose: :+1: I'd be in favor of maintaining 1.0 compatibility, if it's a small check.

mrocklin:
@ian-r-rose is it safe for me to push out a micro release on the Python side?

ian-r-rose:
Yeah, absolutely!

mrocklin:
Had a tiny snafu, but eventually got there.  Version 1.0.3 is on PyPI

PR #90, opened 2019-09-12
[REVIEW] Populate new plots using API request to cluster

### LINKED ISSUE #75: Add new plots through individual-plots.json
The orange buttons in the sidebar are currently hardcoded.  It would be great to have them be defined when we talk to the scheduler.  This would allow us to add new plots in Dask without having to update this package, and to have plots that are only defined in some situations (such as if a GPU is present)

Currently we query `/individual-plots.json` to see if the dashboard is up and running.

```typescript
    // If this is a url that we are proxying under the notebook server,
    // it is easier to check for a valid dashboard.
    if (url.indexOf(settings.baseUrl) === 0) {
      return ServerConnection.makeRequest(
        URLExt.join(url, 'individual-plots.json'),
        {},
        settings
      ).then(response => {
        if (response.status === 200) {
          return true;
        } else {
          return false;
        }
      });
    }
```

I think that originally the plan was to check this to see what plots were available.  Is this still possible?  I'm adding a new plot here (https://github.com/dask/distributed/pull/2922) and suspect adding some more that might only be present sometimes.  

Currently we hard-code these plots here:

https://github.com/dask/dask-labextension/blob/b68342bcfb43f905d7b86588fb3083f917519b5a/src/dashboard.tsx#L417-L426

But we might get them from the individual-plots route

```
$ curl localhost:8787/individual-plots.json
{
  "Individual Task Stream": "/individual-task-stream", 
  "Individual Progress": "/individual-progress",
  "Individual Graph": "/individual-graph", 
  "Individual Profile": "/individual-profile", 
  "Individual Profile Server": "/individual-profile-server", 
  "Individual Nbytes": "/individual-nbytes", 
  "Individual Cpu": "/individual-cpu", 
  "Individual Nprocessing": "/individual-nprocessing", 
  "Individual Workers": "/individual-workers"
}
```

thomcom:
Fixes #75 

Whenever the url changes, request the list of `individual-plots.json` and change `_items` to reflect it.

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -119,10 +119,10 @@ export class DaskDashboardLauncher extends Widget {
     this._input.urlChanged.connect(this.updateLinks, this);
   }
 
-  protected async updateLinks() {
+  private async updateLinks(): Promise<void> {
     const result = await Private.getItems(
       this._input.url,
-      this._input._serverSettings
+      ServerConnection.makeSettings()
     );
     if (result) {
       let newItems: IDashboardItem[] = [];
@@ -372,13 +372,13 @@ export class URLInput extends Widget {
     });
   }
 
-  public _url = '';
-  public _serverSettings: ServerConnection.ISettings;
   private _urlChanged = new Signal<this, URLInput.IChangedArgs>(this);
+  private _url = '';
   private _isValid = false;
   private _input: HTMLInputElement;
   private _poll: Poll;
   private _isDisposed: boolean;
+  private _serverSettings: ServerConnection.ISettings;
 }
 
 /**
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -123,7 +123,7 @@ export class DaskDashboardLauncher extends Widget {
   private async updateLinks(): Promise<void> {
     const result = await Private.getItems(
       this._input.url,
-      ServerConnection.makeSettings()
+      this._serverSettings
     );
     if (result) {
       let newItems: IDashboardItem[] = [];
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -116,7 +116,12 @@ export class DaskDashboardLauncher extends Widget {
     this.addClass('dask-DaskDashboardLauncher');
     this._items = options.items || DaskDashboardLauncher.DEFAULT_ITEMS;
     this._launchItem = options.launchItem;
-    this._input.urlChanged.connect(this.update, this);
+    this._input.urlChanged.connect(this.toot, this);
+    this._startUrlTimer();
+  }
+
+  protected toot(the_arg: any) {
+    this.update();
   }
 
   /**
@@ -159,10 +164,44 @@ export class DaskDashboardLauncher extends Widget {
     this.update();
   }
 
+  /**
+   * Dispose of the resources held by the dashboard.
+   */
+  dispose(): void {
+    this._poll.dispose();
+    super.dispose();
+  }
+
+  /**
+   * Poll for valid urls
+   */
+  private _startUrlTimer(): void {
+    this._poll = new Poll({
+      factory: async () => {
+        const url = this._input._url;
+        const result = await Private.getItems(url, this._input._serverSettings);
+        if (result) {
+          let newItems: IDashboardItem[] = [];
+          for (let key in result) {
+            let label = key.replace('Individual ', '');
+            let route = String(result[key]);
+            let item = { route: route, label: label, key: label };
+            newItems.push(item);
+          }
+          this._items = newItems;
+          this.update();
+        }
+      },
+      frequency: { interval: 4 * 1000, backoff: true, max: 60 * 1000 },
+      standby: 'when-hidden'
+    });
+  }
+
   private _dashboard: Widget;
   private _input: URLInput;
   private _launchItem: (item: IDashboardItem) => void;
   private _items: IDashboardItem[];
+  private _poll: Poll;
 }
 
 /**
@@ -355,12 +394,12 @@ export class URLInput extends Widget {
   }
 
   private _urlChanged = new Signal<this, URLInput.IChangedArgs>(this);
-  private _url = '';
+  public _url = '';
   private _isValid = false;
   private _input: HTMLInputElement;
   private _poll: Poll;
   private _isDisposed: boolean;
-  private _serverSettings: ServerConnection.ISettings;
+  public _serverSettings: ServerConnection.ISettings;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


thomcom:
It's ready, thanks for taking a look!

src/dashboard.tsx
```diff
@@ -116,7 +116,25 @@ export class DaskDashboardLauncher extends Widget {
     this.addClass('dask-DaskDashboardLauncher');
     this._items = options.items || DaskDashboardLauncher.DEFAULT_ITEMS;
     this._launchItem = options.launchItem;
-    this._input.urlChanged.connect(this.update, this);
+    this._input.urlChanged.connect(this.updateLinks, this);
+  }
+
+  protected async updateLinks() {
```
ian-r-rose: Do you envision subclasses making use of this function? If not, I think we should make it private: `private async _updateLinks(): Promise<void>`

src/dashboard.tsx
```diff
@@ -116,7 +116,25 @@ export class DaskDashboardLauncher extends Widget {
...
+    const result = await Private.getItems(
+      this._input.url,
+      this._input._serverSettings
+    );
+    if (result) {
+      let newItems: IDashboardItem[] = [];
+      for (let key in result) {
+        let label = key.replace('Individual ', '');
+        let route = String(result[key]);
+        let item = { route: route, label: label, key: label };
```
ian-r-rose: I am kind of regretting naming this `items`, pretty vague, no?

src/dashboard.tsx
```diff
@@ -354,13 +372,13 @@ export class URLInput extends Widget {
     });
   }
 
+  public _url = '';
```
ian-r-rose: Rather than making these public, let's find out a way to get all the info out that we need. The `URLInput` widget already has a public getter for the URL, so this change isn't necessary. We could also make the dashboard responsible for creating the server connection settings so that it is available to it. Those settings can then be passed in as constructor options to the URL input.

src/dashboard.tsx
```diff
@@ -518,6 +536,30 @@ namespace Private {
...
+  /**
+   * Return the json result of /individual-plots.json
+   */
+  export function getItems(
+    url: string,
+    settings: ServerConnection.ISettings
+  ): Promise<IDashboardItem[] | void> {
+    url = normalizeDashboardUrl(url, settings.baseUrl);
+
+    return ServerConnection.makeRequest(
```
ian-r-rose: This looks good to me. I think it could be made a bit flatter with async/await, but that's just cosmetic.

src/dashboard.tsx
thomcom: How does `buttons` sound, or `links`?

src/dashboard.tsx
thomcom: I'm not experienced enough with Jupyter Labextensions to predict if DaskDashboardLauncher will ever be subclassed. Probably not. :) I'll make that change.

src/dashboard.tsx
thomcom: Can't the `DaskDashboardLauncher` just get its own instance of `ServerConnection.makeSettings()`?

src/dashboard.tsx
thomcom: I copied the `testDaskDashboard` pretty closely. :)

src/dashboard.tsx
thomcom: I had a feeling this would be the stickiest spot. Reverted to old protection status!

src/dashboard.tsx
thomcom: `dashboardLinks`? I'm not sure renaming it is necessary right now honestly.

src/dashboard.tsx
thomcom: 👍

thomcom:
Thanks for your review @ian-r-rose .
As for testing it - I didn't create a different dashboard server with different buttons. I know that it works by monitoring the `updateLinks` callback, though. When the Lab session starts up, it uses the `DEFAULT_ITEMS` list. As soon as it connects to an existing LocalCluster or one created with the +NEW button, the list updates with the result of the `individual-plots` API request. If I click from one LocalCluster to another, the `updateLinks` callback is fired, which changes the button list to the ones provided from that specific `individual-plots` API request.

I don't know how to customize the response to that API request between different servers, but I think it works. :)

src/dashboard.tsx
ian-r-rose: No, I wasn't suggesting you rename these, just musing :)

src/dashboard.tsx
ian-r-rose: It could make its own settings, though in theory I suppose we might want to ensure that the two widgets are using the same server connection settings. I'm unsure when it might be a problem if they were different, however.

src/dashboard.tsx
```diff
@@ -116,7 +116,25 @@ export class DaskDashboardLauncher extends Widget {
...
     this._items = options.items || DaskDashboardLauncher.DEFAULT_ITEMS;
     this._launchItem = options.launchItem;
-    this._input.urlChanged.connect(this.update, this);
+    this._input.urlChanged.connect(this.updateLinks, this);
+  }
+
+  private async updateLinks(): Promise<void> {
+    const result = await Private.getItems(
+      this._input.url,
+      ServerConnection.makeSettings()
```
ian-r-rose: What do you think about making the settings once, keeping them as a private member, and then passing them into the `URLInput` as a constructor arg? I don't think that it's particularly expensive, nor is this function called all that often, but creating them de novo each time rubs be a bit the wrong way.

thomcom: This is your project and I defer to your architectural preferences entirely!

Honestly, from a design standpoint, my preference to not keep a reference to `ServerConnection.makeSettings()` anywhere, but call it any time it is needed. `private _serverSettings: ServerConnection.ISettings` implies that `_serverSettings` is a necessary state variable to `URLInput`, something it will use either as permutable state or as implicit "environment state" (the banana is held by the gorilla). But the settings belong to `ServerConnection`, and should be left to it to coordinate.

thomcom: I made that change as you suggested. Doesn't it seem messy, compared to leaving the `makeSettings()` as the responsibility of `ServerConnection`? Now we've got two different classes with their own private `_serverSettings` objects and potential risk of state bugs relating to them falling out of sync. In an OO context it implies (Please lets not do this!) that there should be a parent object `ServerSettingsStateHolder` that `URLInput` and `DaskDashboardLauncher` both inherit from so that we don't have to dupe all the `_serverSettings` member code.

ian-r-rose: The `ISettings` interface is entirely marked as `readonly`, so at least we should be able to be confident that it won't mutate out from under us: https://github.com/jupyterlab/jupyterlab/blob/1f59d9c1b1e1cae9323e3fa9b3b3a4a98e20ca84/packages/services/src/serverconnection.ts#L53-L101 . I think I also disagree that they belong to `serverConnection`, since they are passed in by the caller of `makeRequest`. I think of `makeSettings` as more of a convenience method for creating default values.

We can make `serverSettings` an optional input for `URLInput`, and it can create its own version if it is not provided. My main goal here is to ensure that the dashboard and the `URLInput` are using the same settings for talking to the server. Right now, we are using the defaults, so they wind up being identical. But if in the future we want to allow that to be customizable, it makes it somewhat easier for them to share a reference to the same settings.

It's not really a big deal either way, though.

thomcom: I've made the changes that allow them to share the result of `DaskDashboardLauncher` requesting the `makeSettings` object. I'm fine to submit it that way if you like it!

mrocklin:
Thanks @thomcom and @ian-r-rose ! 

This unblocks me from making some new dashboard plots.  Previously I wasn't adding them because it wasn't obvious on how to get access to them.  Now that they're more accessible it makes a lot more sense to start investing time here.  I'm looking forward to it!

ian-r-rose:
@mrocklin be advised that there is currently a bug I need to fix that occurs if you try to set the `dashboard_link` to `/proxy/8787/` in your dask config. It should work correctly if you use the internal proxy here (that is, the URLs that are generated by default).

PR #92, opened 2019-09-26
Add conda install jupyterlab nodejs to README

mrocklin:
Getting nodejs installed has been a common speedbump when dealing with
users.
This commit adds a `conda install` line to the installation notes.
We make sure to state that this is just one installation path among many.


ian-r-rose:
Looks good!

PR #96, opened 2019-10-01
Dashboard fallback

### LINKED ISSUE #94: Sidebar no longer detects dashboard
Things in master work well if I create the dashboard from the sidebar, but it no longer seems to work well if I provide the address to an external dashboard

image

Note in this screenshot that the address of the current dashboard is provided to the sidebar, but the lab extension doesn't connect to it.

cc @ian-r-rose @thomcom

ian-r-rose:
Fixes #94 , sort of. I think there is a limit to how good this fix can be -- we can't really make good API requests to the Bokeh server (including the `individual-plots.json` endpoint) from the browser due to CORS restrictions. This is also why we are jumping through some contortions to test whether the endpoint is even there. So this fix just falls back on the old default dashboard settings when a direct URL to a bokeh server is entered. Not ideal, but I wasn't able to find a way around it.

This is tricky to get right because, in addition to trying to work within CORS policies, we are trying to support at least three use-cases for the dashboard URL:
1. A raw bokeh server address (i.e., the case described in #94).
1. A bokeh server proxied using `jupyter-server-proxy` (e.g., `/proxy/8787`)
1. A bokeh server proxied by this extension's cluster manager.

All of which is to say, I'd appreciate another set of eyes on this to see if it works as expected :)

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -121,20 +121,15 @@ export class DaskDashboardLauncher extends Widget {
   }
 
   private async updateLinks(): Promise<void> {
+    if (!this._input.isValid) {
+      this.update();
+      return;
+    }
     const result = await Private.getItems(
       this._input.url,
       this._serverSettings
     );
-    if (result) {
-      let newItems: IDashboardItem[] = [];
-      for (let key in result) {
-        let label = key.replace('Individual ', '');
-        let route = String(result[key]);
-        let item = { route: route, label: label, key: label };
-        newItems.push(item);
-      }
-      this._items = newItems;
-    }
+    this._items = result || DaskDashboardLauncher.DEFAULT_ITEMS;
     this.update();
   }
 
@@ -519,7 +514,7 @@ namespace Private {
    * Optionally remove a `status` route from a dashboard url.
    */
   export function normalizeDashboardUrl(url: string, baseUrl = ''): string {
-    if (URLExt.isLocal(url)) {
+    if (isLocal(url)) {
       if (!baseUrl) {
         baseUrl = PageConfig.getBaseUrl();
       }
@@ -544,25 +539,33 @@ namespace Private {
   /**
    * Return the json result of /individual-plots.json
    */
-  export function getItems(
+  export async function getItems(
     url: string,
     settings: ServerConnection.ISettings
-  ): Promise<IDashboardItem[] | void> {
+  ): Promise<IDashboardItem[]> {
+    // We can't fetch items from a non-local URL due to CORS policies.
+    if (!isLocal(url)) {
+      return DaskDashboardLauncher.DEFAULT_ITEMS;
+    }
     url = normalizeDashboardUrl(url, settings.baseUrl);
-
-    return ServerConnection.makeRequest(
+    const response = await ServerConnection.makeRequest(
       URLExt.join(url, 'individual-plots.json'),
       {},
       settings
-    ).then(response => {
-      if (response.status === 200) {
-        return response.json().then(data => {
-          return data;
-        });
-      } else {
-        return undefined;
+    );
+    if (response.status === 200) {
+      let result = await response.json();
+      const newItems: IDashboardItem[] = [];
+      for (let key in result) {
+        let label = key.replace('Individual ', '');
+        let route = String(result[key]);
+        let item = { route: route, label: label, key: label };
+        newItems.push(item);
       }
-    });
+      return newItems;
+    } else {
+      return [];
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


mrocklin:
Thanks @ian-r-rose .  I'll give this a test drive later today.

mrocklin:
When I click the "New" button I no longer see a cluster show up in the bottom part of the sidebar.

mrocklin:
Ah, wait, that might be my fault.

mrocklin:
But, this is definitely an improvement because I can connect to clusters that weren't created by the sidebar again.

mrocklin:
cc also @jacobtomlinson , who probably understands both the Dask deployment and client-side parts of this.

jacobtomlinson:
Would it be helpful to add CORS headers to the dask dashboard to allow all hosts?

ian-r-rose:
Yes, that would be very helpful @jacobtomlinson, I tried that a while back and was not particularly successful, but I am also not very familiar with the workings of the bokeh server.

mrocklin:
>> Would it be helpful to add CORS headers to the dask dashboard to allow all hosts?

> I tried that a while back and was not particularly successful, but I am also not very familiar with the workings of the bokeh server.

@thomcom , is this something that is easy for you to look into?

mrocklin:
This PR is an improvement over the current state.  Should we merge it in?

ian-r-rose:
Yep, feel free to merge @mrocklin. We can take another crack at the CORS issues in a follow-up. We can also cut a quick release with this, or wait until we have investigated setting CORS headers.

thomcom:
CORS is a protocol that browsers enforce and that clients and servers have to support. If the bokeh server already supports CORS (it does not appear to), then supporting it here is as simple as adding a couple of headers to HTTP requests that the extension originates. If the bokeh server doesn't support CORS, then the harder path becomes making the above fix, and also getting a PR accepted into bokeh to add CORS. A day or two to figure it out/do the easy case, a week to do both.

mrocklin:
> Would it be helpful to add CORS headers to the dask dashboard to allow all hosts?

@jacobtomlinson is this something that you feel comfortable resolving?

jacobtomlinson:
My plan would be to make a configurable option in distributed where the user can set the allowed origins for the dashboard.

Having a quick look through the bokeh and tornado documentation it seems like there isn't an easy way to set server-wide default headers. It must be set on every handler individually.

Given that many of these handlers will be buried within bokeh server logic this does seem to be non-trivial.

@ian-r-rose is there a specific set of URLs you need to talk to, or is it just generally everything?

mrocklin:
> @ian-r-rose is there a specific set of URLs you need to talk to, or is it just generally everything?

I believe that the Dask JLab extension needs to look at the `individual-plots.json` route in order to determine if the scheduler is up, and what buttons to display to the user.

ian-r-rose:
Yeah, I would expect to be able to talk to all of the routes that the dashboard server advertises in `individual-plots.json`.

mrocklin:
@jacobtomlinson checking in here.  Do you have a sense for how long this would take to resolve?

jacobtomlinson:
I think this is one of those issues that requires slower time changes to the upstream library. I haven't had a chance to look at it lately though.

mrocklin:
What specifically are those changes?  I would like to start raising issues and developing a plan here.

ian-r-rose:
If we need to make some upstream changes to the bokeh server, then I think we should probably move forwards with this imperfect solution for now. A longer term fix would likely involve

1. Configuring the bokeh server to support CORS, as described above, and
2. Bringing this extension closer into line with the functionality implemented in `dask-gateway`, possibly merging some aspects for reuse.

mrocklin:
Can I ask someone here who knows more about CORS to raise an issue at https://github.com/bokeh/bokeh ?  They're pretty responsive.  I'm happy to do this as well, but I give myself only about a 50% chance of describing things correctly.

mrocklin:
> Bringing this extension closer into line with the functionality implemented in dask-gateway, possibly merging some aspects for reuse.

cc @jcrist

jacobtomlinson:
Ref bokeh/bokeh#9416

mrocklin:
Thanks Jacob!

mrocklin:
FYI @quasiben

mrocklin:
@ian-r-rose is it appropriate to merge this and release anyway?  I think that it fixes the bug introduced in #94 and my read of this PR is that it doesn't make anything worse.

mrocklin:
@ian-r-rose gave his assent here in side-channel discussion.  Merging.

PR #98, opened 2019-10-03
Support async scale methods

jcrist:
This adds support for `cluster.scale` being an async method. For now
this support is optional, `cluster.scale` can be either an async or sync
method.

See https://github.com/dask/distributed/issues/3107.

### DIFF
--- a/dask_labextension/manager.py
+++ b/dask_labextension/manager.py
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import importlib
+from inspect import isawaitable
 from typing import Any, Dict, List, Union
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


dask_labextension/manager.py
```diff
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import importlib
+from inspect import isawaitable
```
mrocklin: Ah!  I didn't know that this existed.  We could use this in distributed :)

mrocklin:
This looks good to me. Can you verify that things work ok manually?  If so, I'm good to merge.

jcrist:
Yeah, things work fine locally.

Issue #84, opened 2019-08-26
Auto-start-dask doesn't support notebook restarts


mrocklin:
If I engage the "Auto start Dask" option then my notebook gets the `client` state in it automatically.  However, if I then restart my notebook I seem to lose that state.

mrocklin:
@ian-r-rose any thoughts on how one might resolve this?

mrocklin:
Poke @ian-r-rose , I think that @jsignell has some time to work on this this week.  If you have time to  provide a bit of high level guidance on where to start I suspect that we can clear this out.

ian-r-rose:
Sorry for the slow response @mrocklin. I'd be happy to get an extra set of hands on this if you have the time @jsignell.

I have the start of a fix [here](https://github.com/ian-r-rose/dask-labextension/tree/fix-auto-connect-on-restart). The regression here was due to a change in how the websocket connection was handled in kernel restarts due to [this](https://github.com/jupyterlab/jupyterlab/pull/6246) PR.

The tricky thing I ran into when I looked at this last was a race condition between user kernel interactions and the execute request that this extension sends upon restart. This particularly becomes an issue when the user restarts and begins executing cells immediately. We want our client connection code to be injected first, so any cells that depend on it work.

It's also worth noting that JupyterLab 2.0 (in RC right now) has some more changes to how kernel messages are handled (jupyterlab/jupyterlab#7252). I think that those changes might make this problem easier, but we'll probably have to look at that as well.

PR #104, opened 2019-11-26
Add new routes to hard-coded list

mrocklin:
Ideally we would auto-detect these new routes, but for now, let's just
add them manually.


ian-r-rose:
:+1:

Let me finish my thought on fixing #84, and then we can publish a patch release.

PR #105, opened 2019-12-18
Sync versions and add versioneer

jacobtomlinson:
I noticed that the version specified in `package.json` and `dask_labextension/_version.py` were out of sync.

This PR adds versioneer so that the Python package takes its version from the git tags. 

This will enable us to use `jlpm version [--major|--minor|--patch]` command to manage releases going forward. That command will update the version in `package.json` and also create a git tag with the same version, that way both packages should stay in sync.

I'll also look at getting Travis or GitHub Actions to perform the release to PyPI and NPM automatically when the tags are created. I'll put that in a separate PR along with some docs in the readme on how to release, this will be [similar to releases in jupyterlab-nvdashboard](https://github.com/rapidsai/jupyterlab-nvdashboard#publishing).


jacobtomlinson:
Going to merge so I can push on with automating the releases.

PR #110, opened 2020-02-05
Disable "+ NEW" button when cluster is starting

### LINKED ISSUE #103: Visually inform user when we're creating a cluster
When pressing the new cluster button there is currently no feedback to the user that we're doing something.  This is fine if the cluster comes up quickly (as is the case for LocalCluster) but quite stressful when the cluster takes a while (such as is the case when the Scheduler must be run on a remote machine).  In this case the user has the stress of figuring out if they actually pressed the button or not.  If so, they need to wait and not press it again (or else they'll create many clusters).

jsignell:
Closes #103 

After click on "+ NEW", button is disabled and tooltip text changes until cluster is launched:

Screenshot from 2020-02-05 10-48-47

Note that to make this work, I switched from using `ToolbarButton` to `CommandToolbarButton` which allows for easier state updates. This was based on [this](https://gitter.im/jupyterlab/jupyterlab?at=5e399b1c433e1d403995d967) and [this](https://gitter.im/jupyterlab/jupyterlab?at=5e399cf9bfe65274eadcc9c3) comment in gitter.

### DIFF
--- a/src/clusters.tsx
+++ b/src/clusters.tsx
@@ -45,6 +45,8 @@ export class DaskClusterManager extends Widget {
   /**
    * Create a new Dask cluster manager.
    */
+  status = 'ready';
+
   constructor(options: DaskClusterManager.IOptions) {
     super();
     this.addClass('dask-DaskClusterManager');
@@ -112,17 +114,17 @@ export class DaskClusterManager extends Widget {
     );
 
     // Make a shutdown button for the toolbar.
-    toolbar.addItem(
-      'new',
-      new ToolbarButton({
-        iconClassName: 'jp-AddIcon jp-Icon jp-Icon-16',
-        label: 'NEW',
-        onClick: () => {
-          this._launchCluster();
-        },
-        tooltip: 'Start New Dask Cluster'
-      })
-    );
+    const newButton = new ToolbarButton({
+      iconClassName: 'jp-AddIcon jp-Icon jp-Icon-16',
+      label: 'NEW',
+      onClick: () => {
+        this.status = 'starting';
+        this._launchCluster(newButton);
+      },
+      tooltip: 'Start New Dask Cluster',
+      enabled: this.status == 'ready'
+    });
+    toolbar.addItem('new', newButton);
 
     layout.addWidget(toolbar);
     layout.addWidget(this._clusterListing);
@@ -426,7 +428,12 @@ export class DaskClusterManager extends Widget {
   /**
    * Launch a new cluster on the server.
    */
-  private async _launchCluster(): Promise<IClusterModel> {
+  private async _launchCluster(
+    newButton?: ToolbarButton
+  ): Promise<IClusterModel> {
+    if (newButton) {
+      newButton.update();
+    }
     const response = await ServerConnection.makeRequest(
       `${this._serverSettings.baseUrl}dask/clusters`,
       { method: 'PUT' },
@@ -435,10 +442,18 @@ export class DaskClusterManager extends Widget {
     if (response.status !== 200) {
       const err = await response.json();
       void showErrorMessage('Cluster Start Error', err);
+      this.status = 'failed';
+      if (newButton) {
+        newButton.update();
+      }
       throw err;
     }
     const model = (await response.json()) as IClusterModel;
     await this._updateClusterList();
+    this.status = 'ready';
+    if (newButton) {
+      newButton.update();
+    }
     return model;
   }
 
--- a/src/clusters.tsx
+++ b/src/clusters.tsx
@@ -51,9 +51,6 @@ export class DaskClusterManager extends Widget {
   /**
    * Create a new Dask cluster manager.
    */
-  status = 'ready';
-  registry: CommandRegistry;
-  launchClusterId: string;
 
   constructor(options: DaskClusterManager.IOptions) {
     super();
@@ -62,8 +59,8 @@ export class DaskClusterManager extends Widget {
     this._serverSettings = ServerConnection.makeSettings();
     this._injectClientCodeForCluster = options.injectClientCodeForCluster;
     this._getClientCodeForCluster = options.getClientCodeForCluster;
-    this.registry = options.registry;
-    this.launchClusterId = options.launchClusterId;
+    this._registry = options.registry;
+    this._launchClusterId = options.launchClusterId;
 
     // A function to set the active cluster.
     this._setActiveById = (id: string) => {
@@ -169,6 +166,13 @@ export class DaskClusterManager extends Widget {
     IChangedArgs<IClusterModel | undefined>
   > {
     return this._activeClusterChanged;
+  }
+
+  /**
+   * Whether the cluster manager is ready to launch a cluster
+   */
+  get isReady(): boolean {
+    return this._isReady;
   }
 
   /**
@@ -435,8 +439,8 @@ export class DaskClusterManager extends Widget {
    * Launch a new cluster on the server.
    */
   private async _launchCluster(): Promise<IClusterModel> {
-    this.status = 'starting';
-    this.registry.notifyCommandChanged(this.launchClusterId);
+    this._isReady = false;
+    this._registry.notifyCommandChanged(this._launchClusterId);
     const response = await ServerConnection.makeRequest(
       `${this._serverSettings.baseUrl}dask/clusters`,
       { method: 'PUT' },
@@ -445,14 +449,14 @@ export class DaskClusterManager extends Widget {
     if (response.status !== 200) {
       const err = await response.json();
       void showErrorMessage('Cluster Start Error', err);
-      this.status = 'failed';
-      this.registry.notifyCommandChanged(this.launchClusterId);
+      this._isReady = true;
+      this._registry.notifyCommandChanged(this._launchClusterId);
       throw err;
     }
     const model = (await response.json()) as IClusterModel;
     await this._updateClusterList();
-    this.status = 'ready';
-    this.registry.notifyCommandChanged(this.launchClusterId);
+    this._isReady = true;
+    this._registry.notifyCommandChanged(this._launchClusterId);
     return model;
   }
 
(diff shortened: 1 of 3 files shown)


src/clusters.tsx
```diff
@@ -45,6 +46,9 @@ export class DaskClusterManager extends Widget {
   /**
    * Create a new Dask cluster manager.
    */
+  status = 'ready';
+  commands = new CommandRegistry()
```
jsignell: Instead of defining a new one, rewrite https://github.com/dask/dask-labextension/blob/master/src/index.ts#L373-L378

ian-r-rose: The command registry is intended to be a application singleton, so we need to either
1. Pass in the command registry as a constructor option to the cluster manager for use in the toolbar (as I think you are suggesting here), or
2. Abandon the idea of a `CommandToolbarButton`, and use the lower-level approach like [here](https://github.com/jupyterlab/jupyterlab/blob/8289dd8b18a04db3889fb69287b617b2677b6338/packages/notebook/src/default-toolbar.tsx#L66-L87).

It's unfortunate that the regular `ToolbarButton` widget can't update its `enabled` state, I think it's a defect in its design. So until we fix that, I think we need one of the two above workarounds.

ian-r-rose: I have already gone through some contortions to avoid the first option (having to pass in the `CommandRegistry`), see, e.g., [here](https://github.com/dask/dask-labextension/blob/4342a1bf088a8b20aff23c30d4e0d52eb6e10641/src/index.ts#L147-L149)

jsignell: Oops I missed this comment. I don't understand why 1 is undesirable.

jsignell:
@ian-r-rose I switched to passing in the commands in the constructor. If you are not opposed to that, then I can switch the rest of the buttons to use the same pattern.

src/clusters.tsx
ian-r-rose: Well, it's not exactly a problem to pass in the command registry, so much as an effort to separate concerns. My thinking has been this:

The cluster manager should only need to know about things related to the management of dask clusters. If it has a handle to the command registry, then it can access the whole universe of commands that JupyterLab has, and it becomes more difficult to reason about what the cluster manager does, exactly. That is, if I pass in the command registry, it could call any command that JupyterLab has, sometimes in opaque or confusing ways.

Overall, this is not a huge deal, it's largely me being finicky about APIs.

ian-r-rose:
An aside, can you make sure you have the lint/formatting hooks installed? There are a few minor issues that would be caught by that (e.g., missing semicolons, `!=` vs `!==`).

jsignell:
> An aside, can you make sure you have the lint/formatting hooks installed? There are a few minor issues that would be caught by that (e.g., missing semicolons, `!=` vs `!==`).

I was wondering why I was getting a bunch of missing npm warnings. I hadn't activated my eng. Are there no travis tests for lint?

ian-r-rose:
No, I guess I haven't set those up :)

That would also be a welcome PR, but we can run manually for now.

src/clusters.tsx
```diff
@@ -45,13 +51,19 @@ export class DaskClusterManager extends Widget {
   /**
    * Create a new Dask cluster manager.
    */
+  status = 'ready';
```
ian-r-rose: Can you implement this with a getter and a private attribute? That would make it so `status` couldn't be set by third parties. It would also be nice to tighten up the typings here, maybe something like `DaskClusterManager.IStatus = 'ready' | 'starting' | 'failed'`?

src/clusters.tsx
```diff
@@ -45,13 +51,19 @@ export class DaskClusterManager extends Widget {
...
+  registry: CommandRegistry;
+  launchClusterId: string;
```
ian-r-rose: I think these can be private, and moved down with the rest of the private attributes.

src/clusters.tsx
jsignell: No that makes sense. It'd be nice to pass just the one command down to the cluster manager if there needs to be a top level registry. Would you prefer that I use the more react-y approach?

src/clusters.tsx
```diff
@@ -435,10 +445,14 @@ export class DaskClusterManager extends Widget {
     if (response.status !== 200) {
       const err = await response.json();
       void showErrorMessage('Cluster Start Error', err);
+      this.status = 'failed';
```
ian-r-rose: Is there any way to set the status away from 'failed'? I'm not really understanding the semantics: does `ready` mean ready to start a cluster, or that a cluster is ready?

src/index.ts
```diff
@@ -371,9 +373,15 @@ function activate(
 
   // Add a command to launch a new cluster.
   app.commands.addCommand(CommandIDs.launchCluster, {
-    label: 'Launch New Cluster',
-    execute: () => {
-      return sidebar.clusterManager.start();
+    label: 'NEW',
+    execute: () => sidebar.clusterManager.start(),
+    iconClass: 'jp-AddIcon jp-Icon jp-Icon-16',
+    isEnabled: () => sidebar.clusterManager.status != 'starting',
```
ian-r-rose: This should be `!==`. Not sure why the formatter/linter isn't flagging.

src/index.ts
```diff
@@ -371,9 +373,15 @@ function activate(
...
+    caption: () => {
+      if (sidebar.clusterManager.status == 'starting') {
```
ian-r-rose: And here

src/clusters.tsx
ian-r-rose: I have a slight preference for the react-y approach, but if it winds up being too much of a pain, I am happy to take the `CommandRegistry` approach.

src/clusters.tsx
jsignell: Hmm. I guess it should be just two statuses, `ready` to start a cluster and `starting` all of my checks are just for whether status is starting. So I guess I could also make it a bool. Maybe that would be clearner. `starting: bool`

src/clusters.tsx
jsignell: Ok sounds good.

src/clusters.tsx
jsignell: I guess since the command exists in the `CommandRegistry` we should use it in the button. So I think I'll leave it as the command-based approach unless you dislike it too much.

src/clusters.tsx
jsignell: I went with `isReady` so that the bool matches` isEnabled`.

jsignell:
Should I try the react-y approach, or are you good with this @ian-r-rose?

ian-r-rose:
I think I'm okay with this. I haven't had the chance to take it out for a spin yet. I'll try to do that this evening.

jsignell:
Ok, no rush! Just wanted to let you know that I am done working on it unless you want changes :)

src/clusters.tsx
```diff
@@ -111,16 +120,12 @@ export class DaskClusterManager extends Widget {
...
     toolbar.addItem(
-      'new',
-      new ToolbarButton({
-        iconClassName: 'jp-AddIcon jp-Icon jp-Icon-16',
-        label: 'NEW',
-        onClick: () => {
-          this._launchCluster();
-        },
-        tooltip: 'Start New Dask Cluster'
+      this.launchClusterId,
```
ian-r-rose: This variable name needs updating.

src/clusters.tsx
```diff
@@ -111,16 +120,12 @@ export class DaskClusterManager extends Widget {
...
+      new CommandToolbarButton({
+        commands: this.registry,
+        id: this.launchClusterId
```
ian-r-rose: And here.

src/clusters.tsx
```diff
@@ -111,16 +120,12 @@ export class DaskClusterManager extends Widget {
...
-      new ToolbarButton({
-        iconClassName: 'jp-AddIcon jp-Icon jp-Icon-16',
-        label: 'NEW',
-        onClick: () => {
-          this._launchCluster();
-        },
-        tooltip: 'Start New Dask Cluster'
+      this.launchClusterId,
+      new CommandToolbarButton({
+        commands: this.registry,
```
ian-r-rose: And here

src/index.ts
```diff
@@ -371,9 +373,15 @@ function activate(
...
   app.commands.addCommand(CommandIDs.launchCluster, {
-    label: 'Launch New Cluster',
-    execute: () => {
-      return sidebar.clusterManager.start();
+    label: 'NEW',
+    execute: () => sidebar.clusterManager.start(),
+    iconClass: 'jp-AddIcon jp-Icon jp-Icon-16',
+    isEnabled: () => sidebar.clusterManager.isReady,
+    caption: () => {
+      if (sidebar.clusterManager.isReady) {
```
ian-r-rose: This condition is reversed

jsignell:
@ian-r-rose I'm sorry I asked you to review it in such a sloppy state! I have no idea how I didn't notice that it didn't even build.

src/clusters.tsx
jsignell: Yikes.

ian-r-rose:
Looks good to me, thanks @jsignell!

ian-r-rose:
@jacobtomlinson I don't have a lot of experience using versioneer. Can you give me the 20-second version of how to publish patch releases with it?

jacobtomlinson:
Sure!

**TL;DR**

```
jlpm config set version-tag-prefix ""
jlpm version --patch
git push upstream master && git push upstream master --tags
```

I documented the release procedure [here](https://github.com/dask/dask-labextension#publishing).

**Longer explaination**

Versioneer inspects the git tags and log to calculate the current version number. Then when you run `python setup.py sdist bdist_wheel` or similar it replaces the contents of `<package>/_version.py` with a hard coded version instead because you won't be packaging the `.git` directory.

So generically to make a release you run `git tag <version>` to tag it and then build and publish your package.

Because this project has both a JS and Python component I wanted to ensure both the JS and Python components stay in sync. To do that we use `jlpm version` to increment the version because this will both update the `package.json` and also create the git tag. Then you push the tags up to GitHub. 

Travis will then do the `sdist`/`twine`/`npm` uploading bit for you when it sees a new tag.

PR #111, opened 2020-02-05
Lint on travis

jsignell:
Add linting to checks that happen on travis. This came up in https://github.com/dask/dask-labextension/pull/110#issuecomment-582513804


ian-r-rose:
Thanks!

Issue #121, opened 2020-03-19
Update to use Bokeh >=2


andersy005:
It appears that the most recent release of bokeh(v2.0.0) included some breaking changes. As a result, the websocket connection is not being established properly:

<img width="844" alt="Screen Shot 2020-03-19 at 1 55 28 PM" src="https://user-images.githubusercontent.com/13301940/77109358-5f933200-69e9-11ea-96ef-bcb2a2ab8197.png">
<img width="768" alt="Screen Shot 2020-03-19 at 1 54 51 PM" src="https://user-images.githubusercontent.com/13301940/77109363-61f58c00-69e9-11ea-98b6-e9963b111712.png">

I am running:

```
# Name                    Version                   Build  Channel
dask                      2.12.0                     py_0    conda-forge
dask-core                 2.12.0                     py_0    conda-forge
dask-labextension         2.0.0                      py_0    conda-forge
bokeh                     2.0.0            py37hc8dfbb8_0    conda-forge
```

```
$ jupyter labextension list
JupyterLab v2.0.1
Known labextensions:
   app dir: /glade/work/abanihi/softwares/miniconda3/share/jupyter/lab
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
        @jupyterlab/toc v3.0.0  enabled  OK
        @pyviz/jupyterlab_pyviz v0.9.0  enabled  OK
        dask-labextension v2.0.0  enabled  OK
        jupyterlab-nvdashboard v0.1.9  enabled   X

   The following extension are outdated:
        jupyterlab-nvdashboard

   Consider running "jupyter labextension update --all" to check for updates.
```

andersy005:
P.S.: I downgraded bokeh to `v1.4.0`, and everything seems to be working just fine.

mrocklin:
cc @bryevdv in case he has suggestions on where to look

bryevdv:
Bokeh 2.0 was a major number bump with breaking changes. A general migration guide is here:

https://docs.bokeh.org/en/latest/docs/releases.html#release-2-0-0

If you are talking directly to a Bokeh WS handler (was definitely not aware anyone anywhere was doing this) then there will be other adjustments to make. The Bokeh server websocket handler now expects sub-protocol specific + a token.  The best source of truth is this PR:

https://github.com/bokeh/bokeh/pull/9791

Probably most especially the changes here:

[...]

PR #122, opened 2020-03-20
Pin bokeh to less than version 2

jacobtomlinson:
Related to #121 

Pin bokeh to lower than v2 as current installs are broken. Fixes can be addressed later.


xhochy:
Once https://github.com/jupyterhub/jupyter-server-proxy/pull/180 is released, this should be reverted.

PR #123, opened 2020-03-20
Regenerate Yarn lock

jacobtomlinson:
This should resolve current security alerts.


ian-r-rose:
Looks good, thanks @jacobtomlinson

PR #130, opened 2020-05-20
Add docs on enabling extension

jacobtomlinson:
I've noticed in a few situations that the extension needs to be manually enabled after installation. Even on Jupyter versions higher than the 5.2 mentioned in the docs

Therefore I've added the enable line to the install commands. As the command will do nothing if already enabled I see no harm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ following commands:
 ```bash
 pip install dask_labextension
 jupyter labextension install dask-labextension
+jupyter serverextension enable dask_labextension
 ```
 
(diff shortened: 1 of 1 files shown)


ian-r-rose:
Sorry I missed this @jacobtomlinson. The jupyter notebook server uses a `conf.d` approach for automatically picking up extensions. I've noticed that it doesn't seem to work as intended for editable installs, however. Might this be the reason for the extra step that is sometimes needed? Another possibility is that users may have an older version of the server from before the `conf.d` approach was implemented (v5.2, IIRC).

In any event, you are right that it does not hurt to explicitly enable.

Issue #126, opened 2020-04-18
Auto-detect Dashboard URL


msmicker:
Apologies if this is duplicated, but there are nuances to others reporting issues. My jupyterlab and cluster (SGE based) are remote and working splendidly. My network is open so no need for me to worry about port forwarding etc. I can reach the jupyterlab and dask dashboard directly from my workstation browser. The only thing not working is auto-detect. I just need to manually paste the URL that is showing either from the cluster UI or the client connection into the URL box and everything works.

It seems there must be a way for auto-detect to work since I am getting the URL address to paste directly from the extension itself.

mrocklin:
There is a little magnifying glass button next to the URL address bar.  Try
pressing that?

msmicker:
No such luck. I notice it makes the kernel indicator of my notebook blink but the url does not update. It says Dask and a uuid. I delete that and paste the url manually.

mrocklin:
Ok, then I'm not sure and you'll probably have to wait for one of the other
maintainers who knows more here

ian-r-rose:
@msmicker The extension tries to query the address of the dashboard by [executing](https://github.com/dask/dask-labextension/blob/master/src/index.ts#L506) code in the currently active kernel, but it is definitely not bullet proof.

The relevant code block is
```python
try:
    from dask.distributed import default_client as _internal_jlab_default_client
    display(_internal_jlab_default_client().cluster.dashboard_link)
except:
    pass
```
Can you try this manually in your kernel to see what the result is? We may be able to make the snippet more general/robust.

msmicker:
Thanks! This worked.

```python
try:
    from dask.distributed import default_client as _internal_jlab_default_client
    display(_internal_jlab_default_client().dashboard_link)
except:
    pass
```

ian-r-rose:
Thanks for checking @msmicker. Looks like @jacobtomlinson has already done some of the detective work here in https://github.com/dask/distributed/pull/3429. Sounds like we should update to use `Client.dashboard_link` instead of `Client.cluster.dashboard_link`, possibly with a fallback on the prior version for backward compatibility?

jsignell:
@msmicker are you interested in opening a PR for this? It seems that the solution that @ian-r-rose proposes is the right path.

msmicker:
I could try, but probably not in the next ~2-3 weeks? :)

jacobtomlinson:
I think something like this would work.

```python
try:
[...]

PR #141, opened 2020-08-04
Get dashboard link of client, not cluster

mrocklin:
The dask.distributed.Client object already checks a local .cluster field
if it exists.  The approach in this commit is more general and works if
there isn't a local Cluster object.

Fixes https://github.com/dask/dask-labextension/issues/126

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -503,7 +503,7 @@ namespace Private {
   export function checkKernel(
     kernel: Kernel.IKernelConnection
   ): Promise<string> {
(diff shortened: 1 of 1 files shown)


ian-r-rose:
Looks good to me. Are there any backwards compatibility issues here?

mrocklin:
I don't think so, at least not within memory

mrocklin:
Ah, give me a second before merging though.  Experimenting ....

ian-r-rose:
I have an associated fix in #142. If you are satisfied that this doesn't cause backwards compatibility issues, or that we needn't support those versions of distributed, I have no objection to merging this.

mrocklin:
It causes compatibility issues, but only with relatively old versions of Dask.  I'm happy pushing users along.  Merging.

This fix is still helpful for #126 .  We found two unrelated causes of the same symptom I think

PR #142, opened 2020-08-05
Check language in kernelspec

ian-r-rose:
Fixes https://github.com/dask/dask-labextension/issues/126#issuecomment-668892416

The main fix is in `Private.shouldUseKernel`, which I had to make async. I wound up tugging that thread to use some stricter linter settings around forcing all promises to be awaited or explicitly discarded, so there is some noise associated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "scripts": {
     "build": "tsc",
     "clean": "rimraf lib",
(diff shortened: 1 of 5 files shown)


mrocklin:
I can verify that this works well in my case locally.  Thank you @ian-r-rose

ian-r-rose:
No, I think this should be ready to go. One this is in, it's probably worth cutting a release. Probably a major version bump, since #141 is backwards incompatible.

PR #165, opened 2021-01-06
Update Readme: nodejs step - use conda-forge

FabioRosado:
Hello,

I was trying to install the dask-extension to my jupyterlab but it seems that using `conda install nodejs` will install node 10 which is incompatible when we try to build the extension.

The solution seems to be installing nodejs from conda-forge instead (this [stackoverflow question might be helpful](https://stackoverflow.com/questions/62325068/cannot-install-latest-nodejs-using-conda-on-mac)).

Hopefully, this isn't just something that I hit because my conda environment got messed up somehow 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ These are available through a variety of sources.
 One source common to Python users is the conda package manager.
 
 ```bash
-conda install jupyterlab nodejs
+conda install jupyterlab
+conda install -c conda-forge nodejs
 ```
 
 This extension includes both a client-side JupyterLab extension and a server-side
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ distributed >= 1.24.1
 ## Installation
 
 To install the Dask JupyterLab extension you will need both JupyterLab,
-and [Node.js](https://nodejs.org/).
+and [Node.js](https://nodejs.org/), with at least version 12.
 These are available through a variety of sources.
 One source common to Python users is the conda package manager.
 


jrbourbeau (commented):
Thanks @FabioRosado! Do you happen to know what minimum version of `nodejs` is needed? It would  good to include that in the installation instructions as well

cc @jacobtomlinson @ian-r-rose

ian-r-rose:
Thanks @FabioRosado! I think recommending conda-forge is a good thing here. In the past there have been issues with the version in the `default` channel being too old.

> Do you happen to know what minimum version of `nodejs` is needed? It would good to include that in the installation instructions as well

I think any version >10 should be okay. And in a few days we should be able to move this into the developer instructions, as users will no longer need to have node installed at all.

ian-r-rose (approved):

FabioRosado:
I was chatting with Jacob about this and most likely I hit this issue because conda installed jupyterlab version 3. In this version, we will need node > 12.

Jacob told me that you are working on the version 3 support Ian, so maybe this PR isn't relevant at the moment. I've tried to install jupyterlab 2 but I'm having some issues with installing the manager - I keep getting a "not valid npm" package. So I'll have a look at that to see if I can figure out why this is happening 😄

ian-r-rose:
I think saying node >= 12 is fine, it's almost two years old and it can be pretty difficult to track when some deep dependency requires a bumped version of node :/ 

I think this information will still be relevant, we'll just want to move it into the dev instructions.

FabioRosado:
Not sure if this is relevant, but using conda-forge will install the latest node version (15), I'm not sure if this might cause some headaches somewhere?

Would you like me to mention the node >= 12 on the first step or somewhere else?

ian-r-rose:
I'd say we can mention it in the text where node is first described as a dependency.

FabioRosado:
Awesome thank you 😄 👍

jacobtomlinson:
Thanks @FabioRosado!

Issue #166, opened 2021-01-10
Dask labextension not working after upgrade 3.0.0 -> 4.0.1


rabernat:
This is a cross post of https://github.com/pangeo-data/pangeo-docker-images/issues/177

**What happened**:

I pasted a link to a dask gateway cluster dashboard into the dask labextension dialog.

**What you expected to happen**:

I expected that it would work the same as it did in the past, allowing me to open the various windows of the dashboard in jupyterlab.

**Minimal Complete Verifiable Example**:

From jupyterlab
```python
from dask_gateway import Gateway
gateway = Gateway()
cluster = gateway.new_cluster()
display(cluster)
```

Then I copy/paste the dashboad link (for example: <https://hub.binder.pangeo.io/services/dask-gateway/clusters/prod.2103aa32d5a44956b54fe2f3919593b4/status>), into the labextension dialog. Nothing happens. The labextension never becomes active.

![image](https://user-images.githubusercontent.com/1197350/103908821-f17f3f00-50d0-11eb-9243-d804f5cda489.png)

**Anything else we need to know?**:

This works on earlier version of this environment which use dask-labextension 3.0.0.

There are no error messages on the javascript console specifically related to this. I did see these error messages, but I don't think they are related.

```
DevTools failed to load SourceMap: Could not load content for https://hub.binder.pangeo.io/user/pangeo-gallery-default-binder-u2qtxu7g/static/lab/main.cfb9280d54fcb641c354.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE
DevTools failed to load SourceMap: Could not load content for https://hub.binder.pangeo.io/user/pangeo-gallery-default-binder-u2qtxu7g/static/lab/vendors~main.0d37c9e5f02fa65237ee.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE
```

**Environment**:

A complete list of all packages in the environment is [here](https://github.com/pangeo-data/pangeo-docker-images/blob/c8c0d7271be0978518a1058330999dbb888b50d2/pangeo-notebook/packages.txt)

A binder that reproduces this problem can be found here:
https://binder.pangeo.io/v2/gh/pangeo-gallery/default-binder/64d772a/?urlpath=git-pull?repo=https://github.com/pangeo-gallery/physical-oceanography%26amp%3Burlpath=lab/tree/physical-oceanography/01_sea-surface-height.ipynb%3Fautodecode

[...]

PR #167, opened 2021-01-10
Allow errors in redirect check

TomAugspurger:
See https://github.com/pangeo-data/pangeo-docker-images/issues/177 for
context, specifically
https://github.com/pangeo-data/pangeo-docker-images/issues/177#issuecomment-757553103.
This check seemed to be raising an exception on pangeo's clusters,
possibly because of new networking policies adopted by Jupyterhub but we
haven't confirmed that. The upshot is that requests to
`HUB_URL/<cluster>/` return a 404, but `HUB_URL/<cluster>/status` is just fine. IIUC
this request is just made to see if we should follow a redirect. So it
seems OK to ignore exceptions here.

Unfortunately I haven't been able to test this yet.


ian-r-rose:
Thanks for looking into this @TomAugspurger, I presume this is the same issue as #166? Unfortunately, I think this doesn't quite fix the problem. That particular check is new: I've been trying to move more logic onto the server side to mitigate CORS/mixed content issues that have plagued this extension since the beginning. I suspect what is going on is that we are not using the cookies set by the tornado server, and are thus making an unexpectedly unauthenticated request for the dashboard info. This previously was working because the client was making the request.

I'm not sure right now what the best fix should be, I'll have to think about it (if you have any ideas, I'd be glad to hear them).

ian-r-rose:
The new logic (which may need to be revised) is to keep the old client-side check if the base url of the dashboard matches that of the current lab session, otherwise to do the server-side check: https://github.com/dask/dask-labextension/blob/825aab71e2efdc1a547aff8ccce40e383b503bd9/src/dashboard.tsx#L590-L626

The pangeo case doesn't quite fit the client-side condition since the lab session is under `user` and the gateway session is under `services`.

How is the tornado auth for the gateway session handled? This also fits into a wider problem of "How do we teach this extension about auth and access to the dashboards?".

TomAugspurger:
Yes this is the same as #166.

> How is the tornado auth for the gateway session handled?

I may not fully understand the question, but I'll attempt an answer.

Dask Gateway is registered as a JupyterHub service with the prefix `/services/dask-gateway/`.  Authentication is done at https://github.com/dask/dask-gateway/blob/7d89524d0ffd493910c66ef13ac17787ec463a66/dask-gateway/dask_gateway/auth.py#L152-L166. So the header `Authorization: jupyterlab <token>` is included in requests made to the hub by dask-gateway.


I'm not sure if authentication matters here or not. I'm able to make requests to `HUB_URL/services/dask-gateway/<cluster>/individual-plots.json` from a separate browser window. Perhaps something is being cached though.

---

A couple clarifying questions here: We (maybe?) make a client-side request (in the typescript linked above) to check for the dashboard.  That logic is failing since jupyterlab is at `HUB_URL/user/<username>` but the dashboard is at `HUB_URL/services/dask-gateway/<cluster>`.

On the Python side, this may be available at `client.cluster.dashboard_link`. Do you see an easy way to get that information to the javascript side?

TomAugspurger:
> The new logic (which may need to be revised) is to keep the old client-side check if the base url of the dashboard matches that of the current lab session, otherwise to do the server-side check:

Thinking about this again, what's the harm in skipping the check that the base URL matches the notebook session? I don't know exactly what javascript does when there's no return value, but I assume it's falsey, and so we hit https://github.com/dask/dask-labextension/blob/f2735234a30b69118712aae31f313078683e82dd/src/dashboard.tsx#L275-L278.

If we make the request to `<url>/individual-plots.json` anyway, and it fails, we'd return `{url, isActive: false, plots: {}}`. I see checks for `isActive` elsewhere, so perhaps we're already handling it?

scottyhq:
thanks for opening this up @TomAugspurger and exploring solutions. looks like tests are passing, but i'm not sure things are fixed on the pangeo infrastructure. I 've set up a binder config to point to your branch here https://github.com/scottyhq/pangeo-binder-test 

Still seeing errors (in k8s pod logs) when i try to use the labexension like: 
``` 
[W 2021-01-20 22:09:20.172 ServerApp] https://hub.aws-uswest2-binder.pangeo.io/services/dask-gateway/clusters/prod.a091b3d6477341738e4c9434d10236ed/ does not seem to host a dask dashboard
```
but following that link for a new browser window works fine.

TomAugspurger:
Thanks Scott. I'm taking another look at this today.

TomAugspurger:
I have this semi-working now. You can try it at https://binder.pangeo.io/v2/gh/TomAugspurger/pangeo-binder-test/dask-labextension-167.

For reasons I haven't discovered yet, the dask-labextension will sometimes take a full URL like `https://<hub_url>/services/dask-gateway/clusters/<cluster>/status` and transform it according to (I think) `client.dashboard_link`. This might be at https://github.com/dask/dask-labextension/blob/6afe2867dce574d7e7f8cb14d682c5dd5630a69a/src/index.ts#L532.

In the `daskhub` helm chart, we rely on a relative URL for the public address that hosts the dashboard: https://github.com/dask/dask-labextension/blob/6afe2867dce574d7e7f8cb14d682c5dd5630a69a/src/index.ts#L532. So the `dashboard_link` is something like `/services/dask-gateway/clusteres/<cluster>/status`.

The upshot is that dask-gateway needs to be configured to use the absolute URL, rather than a relative URL. Since the pangeo binder already uses the absolute URL things are fine. The pangeo jupyterhubs might need to be updated.

I would gladly take suggestions for how to test this, but I don't see an easy way. Otherwise this should be ready for review if you have a chance @ian-r-rose.

image

TomAugspurger:
Mmm I just had one more thought... Typically someone (I assume in `distributed`) redirects the URL `<dashboard-address>:<dashboard-port>/` to `.../status`. If I do that on a dask-gateway cluster, I instead get a 404. What's different about dask-gateway vs. a LocalCluster?

1. `curl -v dask-gateway-url/`
2. `curl -v lab-url/proxy/8787/`

and look at the difference in output, I see that

- The LocalCluster redirect is a `302`, while the GatewayCluster is a 301.
- The LocalCluster redirect is to `/hub/api/oauth2/...` and includes cookies, and some other things. The GatewayCluster redirect is just to `/status`.

I'll try to see if aligning any of the GatewayCluster behavior fixes that.

TomAugspurger:
So my dask-gateway setup redirects with a 302 to `/status`, which results in an effective url of `'https://HUB_URL/hub/status'`, which of course isn't correct. We want a request for `https://HUB_URL/services/dask-gateway/clusters/<cluster-name>/` to redirect to `https://HUB_URL/services/dask-gateway/clusters/<cluster-name>/status`. I need to figure out exactly who is doing the redirect.

TomAugspurger:
Maybe the right solution is for the Dask Gateway scheduler to be started with a `--dashboard-prefix`. The default is `/`, we would want something like... `/services/dask-gateway/clusters/{cluster.name}/`... But I'm tripped up by is who is inserting the `/hub` here

```python
>>> url1 = f"https://pc-staging.westeurope.cloudapp.azure.com/{gcluster.dashboard_link}".rstrip("status")
>>> r1 = await httpclient.AsyncHTTPClient().fetch(url1, raise_error=False)
>>> r1.effective_url
>>> 'https://pc-staging.westeurope.cloudapp.azure.com/hub/status'
```

that makes it seem like jupyterhub is also redirecting. Checking with this command, that seems to be the case?


```
!curl -sLD - https://pc-staging.westeurope.cloudapp.azure.com/services/dask-gateway/clusters/staging.8334a0096bb4433ea0d346a3896f445c/ -o /dev/null -w '%{url_effective}'

HTTP/2 301 
content-type: text/html; charset=UTF-8
date: Fri, 29 Jan 2021 14:58:50 GMT
location: /status
server: TornadoServer/6.1
strict-transport-security: max-age=15724800
content-length: 0

HTTP/2 302 
access-control-allow-headers: accept, content-type, authorization
content-security-policy: frame-ancestors 'self'; report-uri /hub/security/csp-report
content-type: text/html
date: Fri, 29 Jan 2021 14:58:50 GMT
location: /hub/status
server: TornadoServer/6.1
strict-transport-security: max-age=15724800
x-jupyterhub-version: 1.2.2
content-length: 0

HTTP/2 404 
access-control-allow-headers: accept, content-type, authorization
content-security-policy: frame-ancestors 'self'; report-uri /hub/security/csp-report
content-type: text/html
date: Fri, 29 Jan 2021 14:58:50 GMT
server: TornadoServer/6.1
strict-transport-security: max-age=15724800
x-jupyterhub-version: 1.2.2
content-length: 4836
```

So maybe if we fix the initial redirect from `/status` to the prefixed URL, we won't have the subsequent redirects to `/status`. I'll try that... somehow.

ian-r-rose:
Thanks for keeping on this @TomAugspurger, and sorry for the delays. I should be able to dig in again this weekend.

TomAugspurger:
FYI, I haven't looked into the http prefix for the dask-gateway side of things yet. I likely won't have a chance to in the near-tearm.

scottyhq:
thanks @TomAugspurger and @ian-r-rose for digging into this. Seems like the current implementation works but there is a possibility of simplifying via a change in dask-gateway or jupyterhub config? 

perhaps @consideratio could shed some light on this comment about where the /hub/ redirect is coming from described in this comment https://github.com/dask/dask-labextension/pull/167#issuecomment-769856097

would be awesome to find a fix so that we can level up from labextension=3 !

consideRatio:
> perhaps @consideRatio could shed some light on this comment about where the /hub/ redirect is coming from described in this comment [#167 (comment)](https://github.com/dask/dask-labextension/pull/167#issuecomment-769856097)

It seems like something redirects to /status on the hostname that jupyterhub runs under. That means that the traffic will go to the configurable http proxy, which doesn't have a route for /status but only for /hub/... /users/... /services/... but will default to /hub for unknown routes. This is what cause the second redirect to /hub/status. But, this is not known to jupyterhub, which responds with 404.

The core issue is the initial redirect to /status was absolute, without knowledge of the base prefix of `/services/dask-gateway/clusters/staging.8334a0096bb4433ea0d346a3896f445c/` I think.

dask-gateway is running on the base prefix of `/services/dask-gateway`, while the dask cluster is running under `/services/dask-gateway/clusters/staging.8334a0096bb4433ea0d346a3896f445c/`. In this case it seems like that the spawned dask cluster isn't aware of this and didn't append the base prefix to the /status path.

```
HTTP/2 301 
content-type: text/html; charset=UTF-8
date: Fri, 29 Jan 2021 14:58:50 GMT
location: /status
server: TornadoServer/6.1
strict-transport-security: max-age=15724800
content-length: 0

HTTP/2 302 
access-control-allow-headers: accept, content-type, authorization
content-security-policy: frame-ancestors 'self'; report-uri /hub/security/csp-report
content-type: text/html
date: Fri, 29 Jan 2021 14:58:50 GMT
location: /hub/status
server: TornadoServer/6.1
strict-transport-security: max-age=15724800
x-jupyterhub-version: 1.2.2
content-length: 0

HTTP/2 404 
access-control-allow-headers: accept, content-type, authorization
content-security-policy: frame-ancestors 'self'; report-uri /hub/security/csp-report
content-type: text/html
date: Fri, 29 Jan 2021 14:58:50 GMT
server: TornadoServer/6.1
strict-transport-security: max-age=15724800
x-jupyterhub-version: 1.2.2
content-length: 4836
```

TomAugspurger:
Yep, I think Eric perfectly summarized it.

We could *maybe* fix this in dask-gateway, by ensuring that the dashboard is started with the correct `--http-prefix`, so that it knows what prefix it's being served at. But I still feel like if the user gives dask-labextension the correct full URL, that should just work out of the box (this PR).

roxyboy:
Does dask-labextension v5.0.0 fix this issue...? Trying to start a cluster from the dask dashboard still doesn't seem to work for me.

TomAugspurger:
No. There’s an open PR to fix it, if you want to try it out.

scottyhq:
> if the user gives dask-labextension the correct full URL, that should just work out of the box (this PR). 

@ian-r-rose or @jsignell would it be okay to merge this ? would be a nice fix to have for jupterhub/binderhub + dask-gateway clusters

ian-r-rose:
Sorry to be so slow in getting back to this @TomAugspurger and @scottyhq, and thanks for weighing in on some of the underlying JupyterHub+dask-gateway interactions, @consideRatio.

It would be nice to get some of the base URL prefixing sorted out, but I agree with you @TomAugspurger that if the user provides a valid absolute URL for the dashboard, then that should be honored. In any case, this is a pretty unobjectionable fix, thank you for chasing it down.

I pushed a small tweak to your branch, hope you don't mind. Merging and publishing now to hopefully unblock pangeo/jhub deployments.

ian-r-rose:
Published as `dask-labextension` `5.0.1`

TomAugspurger:
Thanks Ian!

PR #162, opened 2021-01-04
Jlab3

ian-r-rose:
Upgrades this extension to use JupyterLab 3. (superseding #154) This is a pretty major change in the packaging, though the changes to the codebase itself are pretty minor. In particular:

1. The JS assets can now be packaged and distributed using pypi/conda. This is a major improvement over requiring end-users to rebuild the application using webpack + node.
2. Upgrades to TS 4.
3. Replaces the deprecated tslint with eslint

PR #170, opened 2021-01-12
Restore versioneer

ian-r-rose:
Follow-up to #162, restoring versioneer in `setup.py`. @jacobtomlinson I'd appreciate it if you could check if this looks right to you, since you have more experience with versioneer than I.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ distributed >= 1.24.1
 
 ## Installation
 
-To install the Dask JupyterLab extension you will need both JupyterLab,
-and [Node.js](https://nodejs.org/), with at least version 12.
+To install the Dask JupyterLab extension you will need to have JupyterLab installed.
+For JupyterLab < 3.0, you will also need [Node.js](https://nodejs.org/) version >= 12.
 These are available through a variety of sources.
 One source common to Python users is the conda package manager.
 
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
Looks great! I wonder given that JS insists we code the version in `package.json` whether the Python code should read from there instead of using versioneer. There doesn't seem to be an equivalent for the JS community.

ian-r-rose:
Thanks for taking a look @jacobtomlinson! Yeah, reading from the `package.json` is in fact what we were doing before I put versioneer back in place, see ab9d149 .

I also had to do some unfortunate `sys.path` patching in 09d589c because versioneer doesn't quite play nicely with the new PEP 518 style build here. See, e.g., [this](https://github.com/python-versioneer/python-versioneer/issues/193) discussion.

So I think it may be a good idea to look at alternatives for versioneer (maybe `setuptools-scm`?), but I haven't really done the research, and I wanted to just get this working in the short-term again.

jacobtomlinson:
Versioneer has some new maintainers, so if things aren't working correctly it might be worth raising an issue over there.

But I guess my point was, given that we have to manually specify the version in the `package.json` should we bother with auto versioning at all?

PR #171, opened 2021-01-13
Pre commit

ian-r-rose:
Migates linting from `husky` to `pre-commit`, which has better multi-language support, and checks black/flake8 in CI.

This applies `black` for the first time in this codebase, so there is a fair amount of noise associated with that.


jacobtomlinson (approved):
This is really great. Many thanks!

ian-r-rose:
Thanks for the review @jacobtomlinson!

PR #188, opened 2021-05-18
Don't check the root bokeh URL

### LINKED ISSUE #180: Make logs less chatty
Some of the recent work in this extension have made the tornado logs a little overly verbose (including some things which could appear to the user to be errors, but are not). We should cut down on that, and move some things to the debug logger.

### LINKED ISSUE #185: A lot of objects are being created
**What happened**:
Whenever I use the extension after a while I start getting: `distributed.utils_perf - WARNING - full garbage collections took x% CPU time recently (threshold: 10%)`, and the percentage just keeps increasing. 

**What you expected to happen**:
Not getting these warnings.

**Minimal Complete Verifiable Example**:

```python
import gc
from collections import Counter

from dask.distributed import Client

def add(x, y):
    return x + y

def obj_types(n=10):
    return Counter(str(type(obj)) for obj in gc.get_objects()).most_common(n)

client = Client()
```
Link the extension to the created client.

```python
future = client.submit(add, 1, 2)
```
By just running the code above and leaving the notebook idle I can see that the counts for bokeh objects start to rise.
Run 1:
```python
client.run_on_scheduler(obj_types)
# [("<class 'dict'>", 143178),
#  ("<class 'tuple'>", 60317),
#  ("<class 'list'>", 42544),
#  ("<class 'set'>", 41334),
#  ("<class 'function'>", 35093),
#  ("<class 'bokeh.core.property.wrappers.PropertyValueDict'>", 23340),
#  ("<class 'parso.python.tree.Operator'>", 17625),
#  ("<class 'bokeh.core.property.wrappers.PropertyValueList'>", 13784),
#  ("<class 'parso.python.tree.PythonNode'>", 13126),
#  ("<class 'parso.python.tree.Name'>", 12038)]
```
Run 2 (after like 10-15 minutes):
```python
client.run_on_scheduler(obj_types)
# [("<class 'dict'>", 510971),
#  ("<class 'tuple'>", 212323),
#  ("<class 'set'>", 178042),
#  ("<class 'bokeh.core.property.wrappers.PropertyValueDict'>", 106980),
#  ("<class 'bokeh.core.property.wrappers.PropertyValueList'>", 62999),
#  ("<class 'list'>", 59689),
#  ("<class 'function'>", 41300),
#  ("<class 'cell'>", 21047),
#  ("<class 'parso.python.tree.Operator'>", 17625),
#  ("<class 'parso.python.tree.PythonNode'>", 13126)]
```

**Anything else we need to know?**:
Please let me know if I can provide some more insights. I don't have much experience with the `gc` so I'm not sure if that's enough information.

**Environment**:
- Dask version: 2021.2.0
- Python version: 3.8.8
- Operating System: Ubuntu 20.10
- Install method (conda, pip, source): conda

ian-r-rose:
Fixes #185. Checking the root of the dask dashboard triggers a lot of bokeh document creation, which leaks memory. As a workaround, only sniff the known JSON endpoint.

This should be fairly safe, but I'd like to try this on a pangeo JupyterHub to ensure it doesn't break things.

Also fixes #180 while I'm at it

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 21.5b1
     hooks:
       - id: black
         language_version: python3
--- a/dask_labextension/dashboardhandler.py
+++ b/dask_labextension/dashboardhandler.py
@@ -29,28 +29,49 @@ class DaskDashboardCheckHandler(APIHandler):
         try:
             client = httpclient.AsyncHTTPClient()
 
-            # Check the user-provided url
+            # First check for the individual-plots endpoint at user-provided url.
+            # We don't check for the root URL because that can trigger a lot of
+            # object creation in the bokeh document.
             url = _normalize_dashboard_link(parse.unquote(url), self.request)
+            effective_url = None
             individual_plots_url = url_path_join(
                 url,
                 "individual-plots.json",
             )
-            self.log.debug(f"Checking for individual plots at {individual_plots_url}")
-            response = await client.fetch(individual_plots_url)
+            try:
+                self.log.debug(
+                    f"Checking for individual plots at {individual_plots_url}"
+                )
+                individual_plots_response = await client.fetch(individual_plots_url)
+                self.log.debug(f"{individual_plots_response.code}")
+            except httpclient.HTTPError as err:
+                # If we didn't get individual plots, we may have to follow a redirect first.
+                self.log.debug(f"Checking for redirect at {url}")
+                response = await client.fetch(url)
+                effective_url = (
+                    _normalize_dashboard_link(response.effective_url, self.request)
+                    if response.effective_url != url
+                    else None
+                )
+                # If there was no redirect, raise.
+                if not effective_url:
+                    raise err
+
+                individual_plots_url = url_path_join(
+                    effective_url,
+                    "individual-plots.json",
+                )
+                self.log.debug(f"Found redirect at {effective_url}")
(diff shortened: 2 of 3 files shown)


jacobtomlinson (approved):

jrbourbeau:
cc'ing @bryevdv for awareness about the bokeh document creation leaking memory

bryevdv:
@jrbourbeau Bokeh [checks carefully](https://github.com/bokeh/bokeh/blob/55903fc84b5d7e31ec6fd0410558c99e05990749/bokeh/document/document.py#L511-L548) that the module that is created for each session/document only has the expected number of referrers whenever a Document is destroyed. But it's definitely possible for user app code to hold references to things inside the module that outlive sessions/documents, e.g. by creating threads in the app code that do not get cleaned up. I am not ruling out an issue on the Bokeh side of things, but as always, we need an issue with a minimal reproducer to investigate.

ian-r-rose:
Thanks for the references @bryevdv. I was able to reproduce this issue more minimally by polling the dask scheduler dashboard, but not by polling one of the bokeh server examples, so it seems more likely that the proliferation of references is in `distributed`. I would probably be interested in something like the pre-flight check (or even just `HEAD` or `OPTIONS` to see if somebody is listening on the other end), for what it's worth. Or if you have other suggestions for best-practices around talking to a bokeh server I'm happy to hear them.

I'm going to merge and publish this since it can result in a pretty nasty user experience, and suggest we continue the discussion in the linked dask/distributed#4827 (where it will hopefully be less pressing).

PR #200, opened 2021-07-20
Improve button layout

### LINKED ISSUE #198: Remove default dashboard plots
This extension gets its list of dask dashboard plots by querying the dashboard's `individual-plots.json` endpoint. These are then shown to the user as the orange buttons in the extension side bar. If, however, that request to `individual-plots.json` fails, we show a default listing, given by this:

https://github.com/dask/dask-labextension/blob/365913b4f77910e75f5e60114d44f5785b762db7/src/dashboard.tsx#L439-L464

Historically, these default items were useful because we were checking for the `individual-plots.json` from the browser side. This request could fail for a valid dashboard URL (often for CORS-related reasons), and having the fallback meant that we could still iframe useful plots for the user. However, since #159 we are checking for `individual-plots.json` on the server side, which is not subject to the same restrictions as the browser. So as it stands today, the default items are mostly (only?) used to show the greyed out buttons when no dask dashboard is connected.

As the dask dashboard evolves, gains new plots, renames things, etc, it has become annoying to keep the default listing up-to-date. Furthermore, different scheduler plugins or distributions of dask could add different dashboards. So there really isn't much guarantee that the default listing is particularly accurate. We should consider removing them entirely, and just show something like a "No dashboard connected" info box.

Would anybody miss the grey buttons?

### LINKED ISSUE #199: Sort order of plots
Right now they seem to come up in random order and this changes every time

image
image

Probably sorting by alphabetical ordering would be a positive change

ian-r-rose:
Some stylistic updates to the dashboard launcher buttons:

1. Remove default buttons, instead show a "Dashboard not connected" message if there is no valid dashboard (fixes #198).
2. Alphabetize the buttons so they are in a more predictable location (fixes #199 )
3. Changes the layout so buttons are one-per-line (I had changed it in prehistory to layout row-wise with overflow so we could fit more buttons in, but I think that made things too busy, and also made it tougher to find buttons)

Screenshots:
image

image

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -142,7 +142,7 @@ export class DaskDashboardLauncher extends Widget {
     layout.addWidget(this._input);
     layout.addWidget(this._dashboard);
     this.addClass('dask-DaskDashboardLauncher');
-    this._items = options.items || DaskDashboardLauncher.DEFAULT_ITEMS;
+    this._items = options.items || [];
     this._launchItem = options.launchItem;
     this._input.urlInfoChanged.connect(this._updateLinks, this);
   }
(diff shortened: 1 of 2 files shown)


jacobtomlinson (commented):
This looks awesome! Nice work @ian-r-rose.

I wonder if we should provide some kind of tip to the user when the dashboard isn't connected on how to connect one.

ian-r-rose:
Good idea. There’s plenty of room for a more detailed message

ian-r-rose:
Now with a more informative message:
image

jacobtomlinson:
Awesome!

ian-r-rose:
Thanks for the review @jacobtomlinson!

jrbourbeau:
This looks great!

GenevieveBuckley:
Thanks for doing this @ian-r-rose 

I've been using jupyter lab instead of just the notebook more and more these days, and was just coming over here to whinge about exactly this! It's going to be a lot less frustrating now

ian-r-rose:
Published in `dask-labextension==5.1.0`

mrocklin:
Oh man, I am definitely installing that right now

Issue #201, opened 2021-07-20
Remove "Dask " from the titles of tabs


mrocklin:
My JLab panes look like this:

![Screen Shot 2021-07-20 at 6 29 31 AM](https://user-images.githubusercontent.com/306380/126332496-736cb639-97bb-4577-8b4c-593fd4511051.png)

If we removed the word Dask from all of the titles then it would be easier to see what was what.  We already have the icon to distinguish these from other kinds of tabs.

ian-r-rose:
Agreed, the icon should suffice.

A fairly simple fix, requires updating the labels here: https://github.com/dask/dask-labextension/blob/3e35dd277e503e13dfb0950379eab1400abcd729/src/index.ts#L192-L194

and here:
https://github.com/dask/dask-labextension/blob/3e35dd277e503e13dfb0950379eab1400abcd729/src/index.ts#L389

mrocklin:
Should I interpret this last comment as "this is easy so I'll do it soon" or "this is easy so it might be a good opportunity to bring in some new".  If the latter, Alex?

ian-r-rose:
I'm happy to do it soon (it is indeed very easy), but I'm also happy to bring on any extra hands. I suppose I intended it as a flare to see if any interested contributors popped up.

mrocklin:
Go silent masses go!

GenevieveBuckley:
silent masses, lol

Well, I came over here to complain about something (and find out that Ian has *already* fixed it, fantastic), so I may as well fix someone else's frustration while I'm here instead.

PR #205, opened 2021-07-30
Remove 'Dask' from jupyter lab tab headings

GenevieveBuckley:
Closes https://github.com/dask/dask-labextension/issues/201

It's easier to see which tab contains what plot if the headings don't all start with the same four letters (especially for small tabs with limited space).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -189,8 +189,8 @@ async function activate(
       }
 
       // Possibly update the name of the existing dashboard pane.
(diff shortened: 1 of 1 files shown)


ian-r-rose:
Thanks @GenevieveBuckley! CI failure is an unrelated dependency issue which I can fix separately

jrbourbeau:
Thanks @GenevieveBuckley!

PR #220, opened 2021-12-20
Throw in poll factory to trigger backoff behavior.

### LINKED ISSUE #213: Dashboard URL backoff doesn't work right when it gets ConnectionRefusedError
**What happened**:

The lab extension, on Lab startup, is populated with a URL.  In this case it's a proxy URL (via jupyter-server-proxy) but I don't think that matters.  I don't have any clusters running yet.  I would expect it to try the URL, get connection refused (because there is no scheduler endpoint running), and back off from 4 seconds to 60 seconds.  But it didn't: it keeps spamming every four seconds forever.

It looks like this:

```
SingleUserLabApp - ERROR - Uncaught exception GET /nb/user/athornton/proxy/8787/individual-plots.json?1635985458902 (70.176.144.76)
HTTPServerRequest(protocol='https', host='data.lsst.cloud', method='GET', uri='/nb/user/athornton/proxy/8787/individual-plots.json?1635985458902', version='HTTP/1.1', remote_ip='70.176.144.76')
Traceback (most recent call last):
  File "/opt/lsst/software/stack/conda/miniconda3-py38_4.9.2/envs/lsst-scipipe-0.7.0/lib/python3.8/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/opt/lsst/software/stack/conda/miniconda3-py38_4.9.2/envs/lsst-scipipe-0.7.0/lib/python3.8/site-packages/jupyter_server_proxy/websocket.py", line 91, in get
    return await self.http_get(*args, **kwargs)
  File "/opt/lsst/software/stack/conda/miniconda3-py38_4.9.2/envs/lsst-scipipe-0.7.0/lib/python3.8/site-packages/jupyter_server_proxy/handlers.py", line 370, in http_get
    return await self.proxy(port, proxied_path)
  File "/opt/lsst/software/stack/conda/miniconda3-py38_4.9.2/envs/lsst-scipipe-0.7.0/lib/python3.8/site-packages/jupyter_server_proxy/handlers.py", line 230, in proxy
    response = await client.fetch(req, raise_error=False)
ConnectionRefusedError: [Errno 111] Connection refused
```

**What you expected to happen**:

I expected it to back off so my JupyterLab logs are less spammy with connections that aren't going to work.

**Minimal Complete Verifiable Example**:

Set up a dashboard status URL pointing at something that should be proxied via jupyter-server-proxy and watch it keep trying frequently forever.  Once you have a scheduler at the URL, everything is happy.

**Anything else we need to know?**:

My environment is K8s based on Zero to JupyterHub.  JH 1.4, JL 3.2.

**Environment**:

- Dask version: 2021.10.0
- Python version: 3.8.8
- Operating System: Centos 7 (x86_64)
- Install method (conda, pip, source): conda

**Fix suggestion**:

I think in addition to reacting appropriately if you get an HTTP failure code, you need to catch at least `ConnectionRefusedError: [Errno 111] Connection refused` and treat that as if you had gotten an HTTP failure code.

ian-r-rose:
Fixes #213 .

The polling functionality selects a sleep in a random interval, so it doesn't look like a clean exponential backoff, but it should bring down the noise

### DIFF
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -364,7 +364,6 @@ export class URLInput extends Widget {
           urlInfo.url,
           this._serverSettings
         );
-        // Show an error if the connection died.
         if (!result.isActive && urlInfo.isActive) {
           console.warn(
             `The connection to dask dashboard ${urlInfo.url} has been lost`
@@ -377,10 +376,17 @@ export class URLInput extends Widget {
             newValue: result
           });
         }
+        // Throw an error if the connection died. If we don't raise/reject here,
+        // then the poll won't back off.
+        if (!result.isActive) {
+          throw Error(`No connection to ${urlInfo.url}`);
+        }
       },
-      frequency: { interval: 4 * 1000, backoff: true, max: 60 * 1000 },
+      frequency: { interval: 4 * 1000, backoff: true, max: 120 * 1000 },
       standby: 'when-hidden'
     });
+    // When the URL changes, refresh the poll and backoff.
+    this._urlChanged.connect(() => this._poll.refresh(), this);
   }
 
   private _urlChanged = new Signal<this, URLInput.IChangedArgs>(this);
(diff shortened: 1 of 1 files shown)


athornton (approved):
This certainly looks like it will do what I need with respect to actually backing off when there's nothing listening on the other end (specifically the thrown Error).

PR #229, opened 2022-04-29
[WIP] Add extra check for dashboards behind auth

### LINKED ISSUE #190: Authentication for dashboards
### The problem 

Broadly speaking, this extension works by passing around URLs:

1. When a user has a cluster and clicks on the magnifying glass icon, it checks `client.dashboard_link` to find a URL for the dashboard, and uses that to fetch the individual plots endpoints and construct dashboard panes.
2. If a user creates a cluster, then tries to insert client connection code for that cluster into a notebook or text file, it grabs the scheduler address off of the cluster, and gives you a snippet like: `client = distributed.Client(<your-scheduler-address>)`.

This URL-based system has gotten us pretty far, and has been useful for a lot of people. Both of the above work fine if your clusters aren't protected by authentication or other security measures. But if that is not the true (often the case for a cloud or HPC based cluster), then it will fail. That is to say, `Client("tls://some-url:8786")` will not connect without an SSL context, and a dashboard at `https://some-authenticated-service:8787/status` won't connect without authorization headers/tokens.

### Proposal

I don't think we will be able to (or will want to) cover authentication for all the manifold ways in which dask clusters are deployed today. But I think we can add an entrypoints-based plugin system to allow deployers of dask clusters to make sure that their services work with this package (similar to what @jacobtomlinson has done in `dask-ctl`, or what `intake` has done for drivers).

At it's most basic, I'm imagining a simple interface to allow library authors to say whether a given cluster belongs to them, and authenticate with their service as necessary. The basic flow:

#### Connecting to a dashboard

1. Given a dashboard URL, ask an entrypoint provider "does this belong to you?"
2. If yes, pass the provider a tornado request object for a dashboard plot and allow them to do whatever to it (probably add auth headers).
3. Make the request as before.

#### Cluster connection

The `ClusterManager` already sends information about a cluster to the frontend to show in the side pane. So it (or whatever replaces it, cf. #189 ) could
1. Ask an entrypoint provider "does this belong to you?"
2. If so, get a snippet for connecting to the cluster to send to the frontend.
3. That snippet can be inserted in to notebooks/text-files  as before.

###  Dask Gateway

`dask-gateway` also has functionality around authenticating and proxying remote clusters. Is it possible to adopt it as a dependency to handle more than what we do here (see some discussion in #135) ? My instinct is that we don't want to require dask-gateway clusters in all circumstances, and that it won't be able to cover all the use cases we want to be able to handle here (and by "handle" I mean, let plugin authors handle). The converse of having dask-gateway implement a plugin here, on the other hand (fixing #135), should be fairly straightforward. But I'm happy to get pushback on that if others disagree. Especially curious what @rabernat and @scottyhq think.

### LINKED ISSUE #225: Dask Lab-extension not connecting with dask-gateway dashboards
This is a cross-post of https://github.com/Quansight/qhub/issues/1012

## What happened:

I pasted a link to a dask gateway cluster dashboard into the dask lab extension search box, and nothing happened. 

What you expected to happen:

The behavior should be the same one as we can see for `dask.distributed` dashboard links (which does work), e.g. allowing me to open the various windows of the dashboard in Jupiter lab.

## Minimal Complete Verifiable Example:

From jupyterlab
```python
from dask_gateway import Gateway
gateway = Gateway()
cluster = gateway.new_cluster()
```
Then I copy/paste the dashboard link, into the lab extension. Nothing happens. The lab extension never becomes active.

image

## Anything else we need to know?:

This works for `dask.distributed` dashboards.

This is the related response from the DaskDashboardCheckHandler:

`{"url": "https://quansight.qhub.dev/gateway/clusters/dev.d9667a0ca58f4144a68b920838e01650/", "isActive": false, "plots": {}}`

Also, the dashboard is available and we can easily connect to it by the same URI
image

Kernel Environment:
```yaml
  "environment-dask.yaml":
    name: dask
    channels:
      - conda-forge
    dependencies:
      - python
      - ipykernel
      - ipywidgets
      - python-graphviz
      - dask ==2.30.0
      - distributed ==2.30.1
      - dask-gateway ==0.9.0
      - numpy
      - numba
      - pandas
```
A complete environment for the Jupyterlab configuration can be found [here](https://github.com/Quansight/qhub/blob/main/qhub/template/image/jupyterlab/environment.yaml)
Python version: 3.8
Operating System: Linux
Install method: Conda

viniciusdc:
closes #225 

This does not solve #190 completely but presents a second validation step using only a single `fetch` which enables the request to be performed in the same browser session.

There are some points here that I would like to point out:

- I am adding an extra condition here to check for `gateway` in the URL, but that's not ideal. I would like to have another opinion on which mechanic should be used here. I originally thought of passing an extra setting to the lab extension config to enable/disable this optional validation...
- My knowledge of TS/JS is very premature, so I just added this simple branch here. But I think the best approach should have been something like an `Interceptor` to verify the response and then move accordingly to that...

Well, this issue is fixed in our case with these changes, but I would love to know what I can do to improve this contribution and have such a feature added to main :smile:

### DIFF
--- a/schema/plugin.json
+++ b/schema/plugin.json
@@ -23,9 +23,10 @@
     },
     "browserDashboardCheck": {
       "type": "boolean",
-      "title": "Enable Test Dashboard with Fetch",
-      "description": "If set to true, the test dashboard will be validated with the a simple fetch request. This is useful for testing the dashboard when behind authentication.",
+      "title": "Determine whether to validate dashboards via browser check.",
+      "description": "If set to true, the test dashboard will be validated within the browser environment. This is useful for testing the dashboard when behind authentication.",
       "default": false
+    }
   },
   "type": "object"
 }
--- a/schema/plugin.json
+++ b/schema/plugin.json
@@ -24,7 +24,7 @@
     "browserDashboardCheck": {
       "type": "boolean",
       "title": "Determine whether to validate dashboards via browser check.",
-      "description": "If set to true, the test dashboard will be validated within the browser environment. This is useful for testing the dashboard when behind authentication.",
+      "description": "If set to true, the test dashboard will be validated within the browser environment. This is useful for testing the dashboard when behind a browser-cookie based authentication.",
       "default": false
     }
   },
--- a/schema/plugin.json
+++ b/schema/plugin.json
@@ -20,7 +20,12 @@
       "title": "Hide Cluster Manager",
       "description": "Some deployments don't want to or are unable to use the cluster manager feature. Toggle to hide it from the user interface (note: this does not disable the underlying functionality).",
       "default": false
-    }
+    },
+    "enableTestDashboardFetch": {
+      "type": "boolean",
+      "title": "Enable Test Dashboard with Fetch",
+      "description": "If set to true, the test dashboard will be validated with the a simple fetch request. This is useful for testing the dashboard when behind authentication.",
+      "default": false
   },
   "type": "object"
 }
--- a/schema/plugin.json
+++ b/schema/plugin.json
@@ -21,7 +21,7 @@
       "description": "Some deployments don't want to or are unable to use the cluster manager feature. Toggle to hide it from the user interface (note: this does not disable the underlying functionality).",
       "default": false
     },
-    "enableTestDashboardFetch": {
+    "browserDashboardCheck": {
       "type": "boolean",
       "title": "Enable Test Dashboard with Fetch",
       "description": "If set to true, the test dashboard will be validated with the a simple fetch request. This is useful for testing the dashboard when behind authentication.",
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -576,7 +576,9 @@ namespace Private {
   ): Promise<DashboardURLInfo> {
     url = normalizeDashboardUrl(url, settings.baseUrl);
 
+    console.log(url)
     console.log(settings.baseUrl)
+    console.log(settings)
     // If Hostname matches the baseUrl, then we can fetch directly.
     // if (new URL(url).host === settings.baseUrl) {
     //   return fetch(
@@ -607,16 +609,16 @@ namespace Private {
     //     });
     // }
 
+    const test = await fetch(URLExt.join(url, 'individual-plots.json'))
+    console.log(await test.json())
+
     // If this is a url that we are proxying under the notebook server,
     // check for the individual charts directly.
     if (url.indexOf(settings.baseUrl) === 0) {
-      // return ServerConnection.makeRequest(
-      //   URLExt.join(url, 'individual-plots.json'),
-      //   {},
-      //   settings
-      // )
-      return fetch(
-        URLExt.join(url, 'individual-plots.json')
+      return ServerConnection.makeRequest(
+        URLExt.join(url, 'individual-plots.json'),
+        {},
+        settings
         )
         .then(async response => {
           if (response.status === 200) {
@@ -654,6 +656,7 @@ namespace Private {
       settings
     );
     const info = (await response.json()) as DashboardURLInfo;
+    console.log(info)
     return info;
   }
 
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -130,6 +130,7 @@ export class DaskDashboard extends MainAreaWidget<IFrame> {
  * A widget for hosting Dask dashboard launchers.
  */
 export class DaskDashboardLauncher extends Widget {
+  enableDashboardFetch: boolean;
   /**
    * Create a new Dask sidebar.
    */
@@ -138,6 +139,7 @@ export class DaskDashboardLauncher extends Widget {
     let layout = (this.layout = new PanelLayout());
     this._dashboard = new Widget();
     this._serverSettings = ServerConnection.makeSettings();
+    this.enableDashboardFetch = options.enableDashboardFetch;
     this._input = new URLInput(this._serverSettings, options.linkFinder);
     layout.addWidget(this._input);
     layout.addWidget(this._dashboard);
@@ -424,6 +426,10 @@ export namespace DaskDashboardLauncher {
    * Options for the constructor.
    */
   export interface IOptions {
+
+    /** Enable Test Dashboard with Fetch */
+    enableDashboardFetch: boolean;
+
     /**
      * A function that attempts to find a link to
      * a dask bokeh server in the current application
@@ -572,7 +578,8 @@ namespace Private {
    */
   export async function testDaskDashboard(
     url: string,
-    settings: ServerConnection.ISettings
+    settings: ServerConnection.ISettings,
+    enableDashboardFetch: boolean
   ): Promise<DashboardURLInfo> {
     url = normalizeDashboardUrl(url, settings.baseUrl);
 
@@ -609,7 +616,7 @@ namespace Private {
         });
     }
 
-    else if (url.includes("gateway")) {
+    else if (enableDashboardFetch) {
       return fetch(
         URLExt.join(url, 'individual-plots.json')
         )
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -301,9 +301,11 @@ export class URLInput extends Widget {
   /**
    * The in browser dashboard check for authenticated dashboards.
    */
+  get browserDashboardCheck(): boolean {
+    return this.browserDashboardCheck;
+  }
   set browserDashboardCheck(value: boolean) {
     this.browserDashboardCheck = value;
-    this.update();
   }
 
   /**
--- a/src/dashboard.tsx
+++ b/src/dashboard.tsx
@@ -130,7 +130,6 @@ export class DaskDashboard extends MainAreaWidget<IFrame> {
  * A widget for hosting Dask dashboard launchers.
  */
 export class DaskDashboardLauncher extends Widget {
-  enableDashboardFetch: boolean;
   /**
    * Create a new Dask sidebar.
    */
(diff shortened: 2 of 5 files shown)


src/dashboard.tsx
```diff
@@ -609,6 +609,37 @@ namespace Private {
         });
     }
 
+    else if (url.includes("gateway")) {
```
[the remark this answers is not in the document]
viniciusdc: Amazing, thanks for the detailed information @ian-r-rose. I will give it a try on my fork and update this once I got some updates.

src/dashboard.tsx
```diff
@@ -609,6 +609,37 @@ namespace Private {
...
+              return {
+                url,
+                isActive: false,
+                plots: {}
+              };
+            }
+          }
+        )
+        .catch((error) => {
+          console.log('Fetch error: ', error);
```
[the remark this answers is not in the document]
viniciusdc: Sure, I was just debugging and found it interesting to have that information when needed. But thanks for the project information, I will take that out later.

viniciusdc:
Hi @ian-r-rose thanks for the comments above, I tried to add what you suggested, but now sure how to incorporate the changes made to `DaskDashboardLauncher.IOptions` into the `testDaskDashboard` new parameter.

schema/plugin.json
```diff
@@ -20,7 +20,12 @@
       "title": "Hide Cluster Manager",
       "description": "Some deployments don't want to or are unable to use the cluster manager feature. Toggle to hide it from the user interface (note: this does not disable the underlying functionality).",
       "default": false
-    }
+    },
+    "enableTestDashboardFetch": {
+      "type": "boolean",
+      "title": "Enable Test Dashboard with Fetch",
+      "description": "If set to true, the test dashboard will be validated with the a simple fetch request. This is useful for testing the dashboard when behind authentication.",
```
ian-r-rose, line 27: As above, the important thing is that it's a browser-side request, rather than `fetch`. I'm not 100% familiar with the QHub setup, but would it be correct to say that it's useful when a browser cookie is needed for authentication? (rather than just any authentication, which might also happen on the server side)

src/dashboard.tsx
```diff
@@ -609,6 +616,37 @@ namespace Private {
         });
     }
 
+    else if (enableDashboardFetch) {
```
ian-r-rose, line 619: Style nit:
```suggestion
    } else if (enableDashboardFetch) {
```


Note that there are some pre-commit hooks that you can install to run linter/formatting checks automatically. But I can also run them in a separate commit later if you're having trouble with them.

src/index.ts
```diff
@@ -282,6 +282,10 @@ async function activate(
   // or are unable to use it.
   let hideClusterManager: boolean = false;
 
+  // Whether to test the Dask dashboard using a fetch request or to proceed
+  // with default behavior.
+  let enableTestDashboardFetch: boolean = false;
```
ian-r-rose, line 287: Further down the file (~line 345) there is an `onSettingsChanged` callack which is called whenever the setting for this is updated. That's where you would go to the `sidebar` and update it to either make the browser check or the server-side check.

I described sending that data in using an option in the `IOptions`object, but on reflection it could also just be a setter on the object itself. As an example, you might look at the `active` getter/setter on the `DaskDashboard` class.

schema/plugin.json
viniciusdc: I updated the descriptions, that's definitely the case. Thanks for pointing that out.

src/dashboard.tsx
viniciusdc: I ran pre-commit this time, hope these errors disappear :slightly_smiling_face:

viniciusdc:
> Thanks @viniciusdc! Left a few style comments, but I think those could mostly be fixed by getting the pre-commit hooks working.
> 
> Regarding passing the option down to the dashboard check: perhaps my earlier suggestion of using an additional option in an `IOption` object was overkill, we could also just have a setter on the `URLInput` class (since that is the class that actually checks the URL). It's still an unfortunate number of steps, but one fewer than what I suggested before :). So it would look something like
> 
> 1. in `onSettingsChanged` read the new setting value
> 2. Update that value using a new setter on `URLInput`
> 3. The `URLInput` object then passes the setting value into the dashboard check utility function.

Hi @ian-r-rose, many thanks for the explanation. I tried to implement these changes (I think I understand a little bit more how this works :slightly_smiling_face: ). Let me know your thoughts about this. Thanks in advance

src/dashboard.tsx
```diff
@@ -296,6 +298,14 @@ export class URLInput extends Widget {
     return this._urlChanged;
   }
 
+  /**
+   * The in browser dashboard check for authenticated dashboards.
+   */
+  set browserDashboardCheck(value: boolean) {
```
ian-r-rose, line 304: I think you'll need to also add a `get` function here to allow `this.browserDashboardCheck` to work.

Perhaps we should enable [this](https://eslint.org/docs/rules/accessor-pairs) linter rule.

src/dashboard.tsx
```diff
@@ -296,6 +298,14 @@ export class URLInput extends Widget {
...
+    this.browserDashboardCheck = value;
+    this.update();
```
ian-r-rose, line 306: I don't think this `update()` is necessary (though it doesn't do much harm either).

ian-r-rose (commented):
This is looking pretty good @viniciusdc, I just have one request to add `get` accessor, and I think this will be good to go!

src/dashboard.tsx
```diff
@@ -301,9 +301,11 @@ export class URLInput extends Widget {
   /**
    * The in browser dashboard check for authenticated dashboards.
    */
+  get browserDashboardCheck(): boolean {
+    return this.browserDashboardCheck;
```
ian-r-rose, line 305: These should be underscored attributes to not shadow the method names

ian-r-rose: Sorry to be terse! Typically what we do is add a private attribute (see the bottom of this class) which is prefixed with an underscore. It's a bit Java-ey, but hopefully not too bad.

So this line would become
```typescript
return this._browserDashboardCheck;
```
and further down with the other private attributes you would have

```typescript
private _browserDashboardCheck = false;
```

viniciusdc: Cool, just added that. Let me know if its correct :slightly_smiling_face:

ian-r-rose (approved):
This looks great, thanks for all your work on this @viniciusdc! Looks like pre-commit is unhappy, but for a reason unrelated to this PR, so I'll go ahead and fix that myself.

@viniciusdc, can I ask you to confirm that the PR in this state still works for your use-case, since I'm not set up to test it myself?

Also, cc @jsignell, who has requested a similar option in the past.

ian-r-rose:
> where should I modify in the config the new attribute to propagate it?

As a user, you should be able to go to the JupyterLab settings editor (Settings -> Advanced Settings Editor) and find the "Dask" tab. That will open a JSON editor in which you can change the settings. These settings are stored in JSON files on disk (you can find the path to those files if you want to manually edit them by running `jupyter lab paths`. That's probably enough for our purposes here.

As a system admin, you'll want to distribute a settings file to your users that have the appropriate values for these settings. That way they won't need to change the settings themselves. This functionality is documented [here](https://jupyterlab.readthedocs.io/en/stable/user/directories.html#overrides-json).

ian-r-rose:
Whoops, found one more place where we need to pass `this._browserDashboardCheck` to `testDaskDashboard`, mind if I push here @viniciusdc?

viniciusdc:
Hi @ian-r-rose, so I was able to toggle the options and I can see the table 
image
the only problem is that it loses connection right after that 
```
The connection to dask dashboard https://qhubstages2.qhub.dev/gateway/clusters/dev.6f863607b3684dd1b8191cf207877cd2/ has been lost
```

ian-r-rose:
Okay, this looks good to me if you are happy with it @viniciusdc

viniciusdc:
I will check again with this addition, thanks for updating it :slightly_smiling_face:

ian-r-rose:
>  the only problem is that it loses connection right after that

Hmm, weird. Do you see anything in the `Network` tab in your browser dev tools?

viniciusdc:
> > the only problem is that it loses connection right after that
> 
> Hmm, weird. Do you see anything in the `Network` tab in your browser dev tools?

I can see two responses, one from `individual-plots.json` and another one:
```
Request URL: https://qhubstages2.qhub.dev/gateway/clusters/dev.6f863607b3684dd1b8191cf207877cd2/individual-plots.json
Request Method: GET
Status Code: 304 
Remote Address: 35.225.180.207:443
Referrer Policy: strict-origin-when-cross-origin
```
the second one, which contains the expected json, the second one though has the following body
```
isActive: false
plots: {}
url: "https://qhubstages2.qhub.dev/gateway/clusters/dev.6f863607b3684dd1b8191cf207877cd2/"
```

ian-r-rose:
Getting two different responses suggests to me that the issue is with having two different checks to `testDaskDashboard`, one with `true`, one with `false`. Maybe you need to rebuild the extension with my last commit?

viniciusdc:
> Getting two different responses suggests to me that the issue is with having two different checks to `testDaskDashboard`, one with `true`, one with `false`. Maybe you need to rebuild the extension with my last commit?

Sure, I think that might be the case as well, rebuilding now.

viniciusdc:
Hi @ian-r-rose, that did the trick. Thank you, its working flawlessly :smile:

ian-r-rose:
> Hi @ian-r-rose, that did the trick. Thank you, its working flawlessly smile

Great, in it goes, thanks @viniciusdc!

viniciusdc:
Thanks for all the help with this @ian-r-rose!! Hope to contribute more in the future with you all!

ian-r-rose:
Thanks again for this @viniciusdc! I've just published version 5.3.0 of `dask-labextension` which includes this feature. Please let me know if any follow-up work is needed!

PR #248, opened 2022-10-26
Configurable default layout

### LINKED ISSUE #247: Default layout
This might not be a good idea but ...

Often when showing someone the Dask dashboard I ask them to set up a few windows, notably the task stream, progress, and workers memory.  I put these to the side of the notebook.  We even make this recommendation in the tutorial.

Would it make sense to codify this starter template somewhere?  Would it be possible to make a button for example that set these up?  

This is difficult I imagine because it's hard to reason about all of the other panes that may or may not be around.

ian-r-rose:
Fixes #247 
dask-layout


jrbourbeau (commented):
Ah, very cool indeed. @ian-r-rose I'm wondering how can folks make/integrate their own custom layout?

ian-r-rose:
> Ah, very cool indeed. @ian-r-rose I'm wondering how can folks make/integrate their own custom layout?

It still needs documentation, but briefly, you can refer to the names of the `individual-*` plots, and specify how they are going to be put into the main area. So the default looks like this:
```json
        "individual-task-stream": {
          "mode": "split-right",
          "ref": null
        },
        "individual-workers-memory": {
          "mode": "split-bottom",
          "ref": "individual-task-stream"
        },
        "individual-progress": {
          "mode": "split-right",
          "ref": "individual-workers-memory"
        }
```
Which puts the task stream chart on the right, then puts the worker memory chart beneath the task stream chart, then puts the progress chart to the right of the worker memory chart.

You can't make an *arbitrary* layout, since you can't refer to other things like notebooks and terminals, but you can make a little "dask" zone however you like.

jacobtomlinson (commented):
Neat!

ian-r-rose:
I should also note that JupyterLab has recently grown a nice new settings UI that can read json schema, and it works reasonably well for this new configuration object:

image

ian-r-rose:
Also experimenting with a new interaction where the `Client` repr gets a button which automatically populates the dashboard URL and then launches the default layout:
dask-layout-2

The goal being to drastically reduce the number of clicks to get your dashboards setup next to your notebooks.

jrbourbeau (commented):
Thanks @ian-r-rose! I think it would be nice to add some documentation on how users can add there own layouts, but overall I think you should merge this whenever you're happy with it

PR #254, opened 2023-01-09
Fix `flake8` `pre-commit` hook link

jrbourbeau:
We're currently seeing errors like 

```
Run pre-commit/action@v2.0.0
install pre-commit
/opt/hostedtoolcache/Python/3.11.1/x64/bin/pre-commit run --show-diff-on-failure --color=always --all-files
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
An unexpected error has occurred: CalledProcessError: command: ('/usr/bin/git', 'fetch', 'origin', '--tags')
return code: 128
stdout: (none)
stderr:
    fatal: could not read Username for 'https://gitlab.com/': No such device or address
    
Check the log at /home/runner/.cache/pre-commit/pre-commit.log
```

in CI since `flake8` has moved from GitLab to GitHub

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       - id: black
         language_version: python3
         exclude: versioneer.py
-  - repo: https://gitlab.com/pycqa/flake8
+  - repo: https://github.com/PyCQA/flake8
     rev: 3.8.3
     hooks:
       - id: flake8


jrbourbeau (commented):
Merging as this change seems pretty uncontroversial and gets CI passing again

PR #258, opened 2023-02-18
Append any URL query args to dashboard URL paths

ntabris:
If you enter `http://1.2.3.4/status?token=abc` as the dashboard address, the existing code will attempt to hit `http://1.2.3.4/status?token=abc/individual-plots.json`.

This makes it instead hit `http://1.2.3.4/individual-plots.json?token=abc`, and also append `?token=abc` to the paths of the individual plots.

With this change, it's possible to use dask-labextension with a dashboard that has token-based auth. (We're now doing this in Coiled, using an NGINX sidecar in front of the dashboard.)

image

### DIFF
--- a/dask_labextension/dashboardhandler.py
+++ b/dask_labextension/dashboardhandler.py
@@ -29,6 +29,14 @@ class DaskDashboardCheckHandler(APIHandler):
         try:
             client = httpclient.AsyncHTTPClient()
 
+            # Extract query (if any) from URL, this will then be appended after path.
+            # This allows using (eg) "?token=[...]" in URL for authentication.
+            if "?" in url:
+                pos = url.find("?")
+                url, query = url[:pos], url[pos:]
+            else:
+                query = ""
+
             # First check for the individual-plots endpoint at user-provided url.
             # We don't check for the root URL because that can trigger a lot of
             # object creation in the bokeh document.
@@ -36,7 +44,7 @@ class DaskDashboardCheckHandler(APIHandler):
             effective_url = None
             individual_plots_url = url_path_join(
                 url,
-                "individual-plots.json",
+                f"individual-plots.json{query}",
             )
             try:
                 self.log.debug(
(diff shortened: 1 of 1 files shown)


dask_labextension/dashboardhandler.py
```diff
@@ -73,6 +81,13 @@ async def get(self, url) -> None:
 
             individual_plots = json.loads(individual_plots_response.body)
 
+            # If there was query in original URL, append to URLs returned
+            if query:
+                for name, plot_url in individual_plots.items():
+                    individual_plots[name] = f"{plot_url}{query}"
+                url = f"{url}{query}"
+                effective_url = f"{url}{query}"
```
ian-r-rose, line 89: I don't think this logic for `effective_url` is correct:

1. `effective_url` could be different from `url` if there is a redirect in the initial `url` (some deployments have had that in the past, and we needed to follow the redirect to get the right individual plot urls).
2. `url` is updated on the previous line, so this *always* appends the query twice to `effective_url`, which you can confirm by inspecting the network tab.

So I think the correct logic would be something like:

```python
    if effective_url:
        effective_url = f"{effective_url}{query}"
```

ian-r-rose (commented):
Thanks @ntabris! Yes, I was a bit hasty in my comment, and (1) isn't necessary for this use-case.

One minor fix, but I think this otherwise looks good.

dask_labextension/dashboardhandler.py
ntabris: Yep, thanks! That was entirely a copy-paste error.

ntabris:
I'll note that loading static assets is requiring some (yet to be determined) shenanigans on our side.

Since the dashboard panes load in iframe, cookies don't work unless the user enables third-party cookies. Getting the auth token in the referer works in Chrome but not Safari, since Safari intentionally doesn't pass along full referer (again, unless you explicitly disable privacy setting).

Suggestions welcome if anything comes to mind.

ian-r-rose:
> 

Hmm, nothing comes to mind immediately. How does Firefox do?

ian-r-rose:
Thanks @ntabris!

ian-r-rose:
I've published this in `6.1.0`

ntabris:
Thanks, much appreciated!

PR #263, opened 2023-06-15
Upgrade extension for JLab 4

mahendrapaipuri:
- This PR upgrades the node dependencies of `@jupyterlab` to `4.0.0` and `@lumino` to `2.0.0`. 
- Minor corrections in `index.ts` are made to work with JLab 4
- Migrate to [hatch](https://hatch.pypa.io/latest/) build system which is now recommended build system by Jupyter project.
- Migrate from `versioneer` to `nodejs` versioning. Python package will take the version string from `package.json`. Bumping of versions can be done using [`hatch version`](https://hatch.pypa.io/latest/version/) command. 
- Added `black` and `pytest` config to `pyproject.toml`
- Use JLab 4 as build and run time dependencies
- Update min Python version to `3.8` inline with JLab 4
- Only JLab 4 is included in CI test cases. Python versions are also bumped to `3.8` and `3.11` in CI
- Use actions from [jupyterlab-maintainer-tools](https://github.com/jupyterlab/maintainer-tools) that does the setting up of Python and nodejs

Maintainers - If you prefer to have `versioneer`, let me know. I will add it back. The thing is `versioneer` is not very well integrated with `pyproject.toml`. There is [hatch-vcs](https://github.com/ofek/hatch-vcs) plugin that is made for hatch build system that does exactly same thing as versioneer but with less config burden.


jacobtomlinson (commented):
Thanks you so much for raising this.

Did you follow the [migration guide](https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html) to generate this? I've run the tool myself and it seems to have made many more changes than are included in this PR. However I also can't get things to install correctly after running the tool.

mahendrapaipuri:
@jacobtomlinson Not really. I just bumped the node deps manually to `4.0.0`. The script that you mentioned will migrate the extension to [hatch](https://hatch.pypa.io/latest/build/) build system which is recommended way to build extensions. Folks at Jupyter developed a [hatch-jupyter-builder](https://github.com/jupyterlab/hatch-jupyter-builder) plugin for the hatch build system which avoids all the [custom logic](https://github.com/dask/dask-labextension/blob/46ad95810ddd947c9fd8f16534835b6a5e32f36d/setup.py#L8-L66) in the `setup.py` file.

If you want, I can extend this PR to hatch migration. We have already done it for our custom extensions and we have a very positive experience.

jacobtomlinson:
If you're happy to take that on that would be fantastic.

mahendrapaipuri:
@jacobtomlinson Please check the updated PR description!

ian-r-rose:
@mahendrapaipuri thank you so much for this, and apologies for the slow review! I'm prioritizing this to get it over the line, and will have some more detailed comments this evening.

We just published a patch release targeting JupyterLab 3, and there are a few conflicts to resolve. I'm happy to do that unless you get to it faster.

I'm very excited to see versioneer and the `setup.py` go away :)

mahendrapaipuri:
@ian-r-rose You are welcome. I dont see any huge problems due to conflicts that are raised due to #267. Most of the conflicting files (`MANIFEST.in`, `yarn.lock` and `setup.py`) will be obsolete and I will update the `package.json` will the new packages that you added in the #267.

ian-r-rose (commented):
Thanks @mahendrapaipuri. I'm still digesting the hatch-based changes, but had a few questions/comments

pyproject.toml
```diff
@@ -1,3 +1,132 @@
...
+    "jupyter-server-proxy>=1.3.2",
+    "jupyterlab>=4.0.0",
+]
+dynamic = ["version", "description", "authors", "urls", "keywords"]
+
+[project.optional-dependencies]
+test = [
+    "pytest",
+    "pytest-cov",
+    "pytest-html",
```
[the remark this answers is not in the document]
ian-r-rose: Ah, never mind, I see them below. Seems reasonable to me.

ian-r-rose (commented):
Overall, this looks really great, my comments and questions are all pretty minor. Thank you @mahendrapaipuri!

.pre-commit-config.yaml
```diff
@@ -1,22 +1,20 @@
...
   - repo: https://github.com/PyCQA/flake8
-    rev: 3.8.3
+    rev: 6.0.0
     hooks:
       - id: flake8
         language_version: python3
-        exclude: versioneer.py
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.1.1
+    rev: v3.0.0-alpha.9-for-vscode
```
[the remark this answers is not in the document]
mahendrapaipuri: I was bumping major versions of all dependencies and back then there was not a final release. Now that I see there is [v3.0.0](https://github.com/pre-commit/mirrors-prettier/tree/v3.0.0) release, I will change it.

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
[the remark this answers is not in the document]
mahendrapaipuri: This is new [config file](https://yarnpkg.com/configuration/yarnrc) in Yarn 3 (which JupyterLab 4 uses). And now Yarn creates a local cache as well at `.yarn` to keep downloaded packages and here we are saying to use `node_modules` to build the package.

setup.py
```diff
@@ -1,107 +1 @@
...
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
-        "Framework :: Jupyter",
-    ],
-)
-
-
-if __name__ == "__main__":
-    setuptools.setup(**setup_args)
+# this file intentionally left blank for legacy tools to find
```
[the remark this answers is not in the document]
mahendrapaipuri: I think this was done for making editable installs work for the projects using `setuptools` as build system. Seems like it is supported since [v64.0.0](https://setuptools.pypa.io/en/latest/history.html#v64-0-0). In any case, we are using `hatch` here and it already supports editable installs and I agree that we can safely remove it.

pyproject.toml
```diff
@@ -1,3 +1,132 @@
...
+[tool.jupyter-releaser.options]
+version_cmd = "hatch version"
+
+[tool.jupyter-releaser.hooks]
+before-build-npm = [
+    "python -m pip install 'jupyterlab>=4.0.0,<5'",
+    "jlpm",
+    "jlpm build:prod"
+]
+before-build-python = ["jlpm clean:all"]
```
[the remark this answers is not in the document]
mahendrapaipuri: The idea was to discuss with the maintainers about migrating to `jupyter-releaser`. 

Ok, understood! I will remove this config for this PR as it will be not used for the time being.

mahendrapaipuri:
@ian-r-rose Addressed your PR comments, rebased onto new main and updated README too.

README.md
```diff
@@ -43,18 +44,25 @@ pip install dask-labextension
...
+For JupyterLab 3.x, use lastest supported version `6.2.0`.
+
+```bash
+pip install dask-labextension==6.2.0
+```
+
+### JupyterLab 2.x
+
+Prior to JupyterLab 3.0 client-side and server-side components needed 
+to be installed separately,nwith node available on the machine.
```
ian-r-rose, line 56:
```suggestion
to be installed separately, with node available on the machine.
```

README.md
```diff
@@ -218,18 +226,31 @@ and a back-end component written in Python.
...
-
 ### Release process
 
-This requires `node`, `build`, and `twine` to be installed.
+This requires `node`, `build`, `hatch` and `twine` to be installed.
 
 ```bash
-jlpm version [--major|--minor|--patch]  # updates package.json and creates git commit and tag
+# To set version (e.g. 7.0.0). hatch will update version string in package.json 
+hatch version "7.0.0"
```
ian-r-rose, line 235: I don't love that this rewrites the `package.json`, but I suppose it will mostly just cause a large diff the first time. Thanks for updating the build docs!

src/index.ts
```diff
@@ -730,10 +731,8 @@ client = Client()`;
     cluster: IClusterModel,
     editor: CodeEditor.IEditor
   ): void {
-    const cursor = editor.getCursorPosition();
-    const offset = editor.getOffsetAt(cursor);
     const code = getClientCode(cluster);
-    editor.model.value.insert(offset, code);
+    editor.model.sharedModel.setSource(code);
```
ian-r-rose, line 735: I need to familiarize myself a bit more with ydoc, but it seems like [`updateSource`](https://jupyter-ydoc.readthedocs.io/en/latest/api/classes/YBaseCell.html#updateSource) would be less destructive if there is  existing text in a cell.

ian-r-rose (approved):
Thanks @mahendrapaipuri! One small suggestion, which I could do in a follow-up if you feel like wrapping this one up.

README.md
mahendrapaipuri: Well, it will update _only_ the version in `package.json`. We leverage the `hatch version` command to bump the version in `package.json` instead of doing it manually. I am not sure what do you mean by large diff.

src/index.ts
mahendrapaipuri: Makes sense. I will test it before making the change!

README.md
ian-r-rose: What I meant was that the hatch-nodejs-version plugin re-orders a bunch of the keys of the `package.json` when bumping the version (at least it did when I tried it yesterday). But that will only really produce a large diff on the first time we use it. I don't think it's a big deal, just a minor annoyance.

src/index.ts
mahendrapaipuri: We are always [inserting a new cell](https://github.com/dask/dask-labextension/blob/73c45ce5f726dbdac5bcfa53ed99cee5f8ab1256/src/index.ts#L792) to inject this code. So, `updateSource` would not make a lot of difference.

Its good that you pointed it out. Seems like we need to wait for the cell to be ready before editor is available. Probably this is JLab 4 change too. I have fixed it but if you think we can do it better, please let me know.

ian-r-rose: > We are always [inserting a new cell](https://github.com/dask/dask-labextension/blob/73c45ce5f726dbdac5bcfa53ed99cee5f8ab1256/src/index.ts#L792) to inject this code. So, updateSource would not make a lot of difference.

Interesting, I had forgotten that, but right you are! https://github.com/dask/dask-labextension/commit/a77b0b45ec3457586e2f55ab2030ad81e556ffb9

In the case of a console or code editor we are not doing that, but I don't know if too many people are doing that in practice, so I'm not too worried about it in practice. If anybody complains, it's easy enough to fix.

ian-r-rose (approved):
Thanks again for all your work on this! I'm going to try to get a release out today

PR #272, opened 2025-02-07
simplify the url pattern for cluster_id

minrk:
rather than requiring it to look like a UUID, accept any single URL path component

results in clearer error: "no such cluster" instead of generic 404 with no matching router and allows for alternate id patterns (e.g. Gateway clusters, which already have ids)


jacobtomlinson (approved):
Thanks @minrk

Issue #204, opened 2021-07-28
Listing DaskGateway clusters created via Python code alongside those created via the dask labextension UI


consideRatio:
**What happened**:

I can create a dask-gateway cluster via the dask-labextension view and I'll see it visible there then.

![starting-new-dask-cluster](https://user-images.githubusercontent.com/3837114/127248817-57492f22-223a-4028-8736-eb8745e90cc3.gif)

But, if I create a dask-gateway cluster from a notebook using code like below, then no dask cluster shows up in the list of clusters.

```python
from dask_gateway import Gateway
gateway = Gateway()
cluster = gateway.new_cluster()
```

**My wish**

My wish is that the dask clusters I've created should be listed visually. I'm not sure if this is possible or not, but I'd like to describe this wish here to explore if we can make it happen one way or another.

**Environment**:

JupyterHub (1.1.1 Helm chart) + Dask-Gateway (0.9.0 Helm chart).

```
$ conda list | grep dask
dask                      2021.6.0           pyhd8ed1ab_0    conda-forge
dask-core                 2021.6.0           pyhd8ed1ab_0    conda-forge
dask-gateway              0.9.0            py38h578d9bd_0    conda-forge
dask-glm                  0.2.0                      py_1    conda-forge
dask-kubernetes           2021.3.1           pyhd8ed1ab_0    conda-forge
dask-labextension         5.0.2              pyhd8ed1ab_0    conda-forge
dask-ml                   1.9.0              pyhd8ed1ab_0    conda-forge
pangeo-dask               2021.06.05           hd8ed1ab_0    conda-forge
```

```
$ python --version
Python 3.8.10
```

Operating System: Ubuntu 20.04
Install method: conda-forge

```
# The current environment and dask configuration via environment
DASK_DISTRIBUTED__DASHBOARD_LINK=/user/{JUPYTERHUB_USER}/proxy/{port}/status
DASK_GATEWAY__ADDRESS=http://10.100.116.39:8000/services/dask-gateway/
DASK_GATEWAY__AUTH__TYPE=jupyterhub
DASK_GATEWAY__CLUSTER__OPTIONS__IMAGE={JUPYTER_IMAGE_SPEC}
DASK_GATEWAY__PROXY_ADDRESS=gateway://traefik-prod-dask-gateway.prod:80
DASK_GATEWAY__PUBLIC_ADDRESS=/services/dask-gateway/
DASK_LABEXTENSION__FACTORY__CLASS=GatewayCluster
[...]

PR #273, opened 2025-02-18
make all manager methods async

minrk:
consistent API makes addressing #204 easier. I ran into issues because my approach to `list_clusters` relies on making a request to the Gateway to list clusters, which should be async, instead of relying on in-memory state which may be out-of-date. I also found it simpler to avoid holding open connections to all clusters, instead only connecting when information is requested, so the Manager is essentially stateless when working with a Gateway.
Gateway clusters also have a 'pending' state, which local clusters don't seem to, which is important to consider, since connecting to a pending cluster will not succeed until/unless the cluster starts.

Also moves the global `manager` state from module global to initialized at extension load time, which works better when instantiating an asyncio task, which may not connect to the right event loop.

### DIFF
--- a/dask_labextension/__init__.py
+++ b/dask_labextension/__init__.py
@@ -2,12 +2,13 @@
 
 from jupyter_server.utils import url_path_join
 
-from . import config
+from . import config  # noqa
 from .clusterhandler import DaskClusterHandler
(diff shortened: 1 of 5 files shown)


jacobtomlinson (approved):
This seems like a good improvement to me! Thanks!

PR #275, opened 2025-05-30
fix: Stop lumino polling for clusters when connection is lost

mahendrapaipuri:
* Currently lumino polls for available clusters with a given frequency. However, when the server terminates, the polling does not stop as long as the browser tab is active. This is an issue in JupyterHub deployments where terminated servers keep making requests to hub when the browser tabs are kept open. This is undesirable as hub still needs to respond to these requests although they end up with 424.
* More background on [JupyterLab #17567](https://github.com/jupyterlab/jupyterlab/pull/17567)

@jacobtomlinson @ian-r-rose  Could you please take a look at it? Also should we drop support for Python 3.8 and add support for 3.12 and 3.13 which will be inline with JupyterLab? Cheers!!

### DIFF
--- a/src/clusters.tsx
+++ b/src/clusters.tsx
@@ -146,7 +146,7 @@ export class DaskClusterManager extends Widget {
         await this._updateClusterList();
       },
       frequency: { interval: REFRESH_INTERVAL, backoff: true, max: 60 * 1000 },
-      standby: 'when-hidden'
+      standby: options.refreshStandby || 'when-hidden'
     });
   }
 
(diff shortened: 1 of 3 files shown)


mahendrapaipuri:
I think the [failed CI](https://github.com/dask/dask-labextension/actions/runs/15343575190/job/43174831819?pr=275) workflow is due to `hatch` not supporting 3.8

jacobtomlinson (approved):
I'm a little out of my depth with the changes here, but overall it makes sense. If you've tested it and can confirm that the desired functionality has been fixed then I'm happy to merge this.

Also yes I agree we should bump up our CI versions here.

mahendrapaipuri:
> I'm a little out of my depth with the changes here, but overall it makes sense. If you've tested it and can confirm that the desired functionality has been fixed then I'm happy to merge this.

Fair enough. As I included in the PR description, we did the same fix for few extensions on the upstream JupyterLab. Yes, I have tested it with JupyuterHub and I confirm that the browser puts polling in stand-by once connection to the single user server is lost.

I have rebased the branch onto the main with #276

jacobtomlinson (approved):
Much appreciated

PR #276, opened 2025-06-02
Drop unsupported Python versions and update action workflows

mahendrapaipuri:
* Unsupported Python versions have been dropped from `pyproject.toml` and minimum supported Python has been bumped to `3.10` in line [dask](https://github.com/dask/dask/blob/87a177d866521a573f7a147cbee4dec7235f88f4/pyproject.toml#L29)
* CI workflows have been updated and action versions have been bumped
* Linters versions have been bumped in the pre-commit config

### DIFF
--- a/.github/workflows/labextension.yml
+++ b/.github/workflows/labextension.yml
@@ -12,12 +12,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.11']
+        python-version: ['3.9', '3.13']
         jupyterlab-version: ['4']
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Base Setup
         uses: jupyterlab/maintainer-tools/.github/actions/base-setup@v1
       - name: Install dependencies
--- a/.github/workflows/labextension.yml
+++ b/.github/workflows/labextension.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


.github/workflows/labextension.yml
```diff
@@ -12,12 +12,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.11']
+        python-version: ['3.9', '3.13']
```
jacobtomlinson: Dask projects only support Python 3.10 and above. Is there any benefit to still supporting 3.9 here?

https://github.com/dask/dask/blob/87a177d866521a573f7a147cbee4dec7235f88f4/pyproject.toml#L29

jacobtomlinson (commented):
Thanks for jumping on this!

.github/workflows/labextension.yml
mahendrapaipuri: Not really!! I just used upstream JupyterLab as the reference as its lowest supported Python version is `3.9`. But I think if dask does not support it, I agree we should bump it to `3.10` directly.

mahendrapaipuri:
@jacobtomlinson CI went green!! Updated the [PR description](https://github.com/dask/dask-labextension/pull/276#issue-3109736275) as well. Let me know your feedback. Cheers!

jacobtomlinson (approved):

mahendrapaipuri:
@jacobtomlinson Do you think doing a minor release with these changes makes sense? Seems like last stable release was in Nov 2023.

